PR #7, opened 2013-08-09
Infer Toggl project from Asana project name, if possible.

vshih:
Also infer Toggl task (for pro workspaces) from Asana task name prefix, if possible.
Calculate time-entry parameters on timer click rather than drop-down update.

### DIFF
--- a/scripts/content/asana.js
+++ b/scripts/content/asana.js
@@ -54,6 +54,7 @@
 
   function addButton(e) {
     if (!(e.target.className === "details-pane-redesign" || iframeRegex.test(e.target.name))) { return; }
+		if ($(".toggl-select")) { return; }
 
     var taskDescription = $(".property.description"),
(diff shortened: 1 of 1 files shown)


Saynah:
The project field on the toggle button is still not being populated for me. When I click the project selector all I get is "no project" (see attached).
It should say "Per Exercise Metrics"

image

vshih:
Does the project already exist in Toggl?  This is one requirement for this to work.  From the project drop-down you are showing, it looks like you don't have any projects defined in Toggl yet.

Saynah:
After closing my browser and reopening, it worked.
However, I still cannot seem to enter tags.

-Mark
(sent from my phone)

IndrekV:
From Toggl Button the only new info you can enter is description. All other data like projects, tags, tasks have to be already entered from toggl.com web or one of the other apps. You can only choose other pre entered data from the Toggl Button form.

PR #184, opened 2014-12-19
Adding TFS

ReddwarfCro:
I use Team Foundation Server and i really missed Toggl-button extension, so i extended this extension. I created TFS.js in witch is all code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Add Toggl one-click time tracking to popular web tools.
   - [Trac][25]
   - [Wunderlist][26]
   - [Toodledo][27]
-  - [TFS (visualStudioOnline and local instalations)][28]
+  - [TFS (visualStudioOnline)][28]
 
 ## Installing from the Web Store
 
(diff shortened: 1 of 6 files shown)


IndrekV:
The permission "*://*/tfs/DefaultCollection/*" requires the user to give permissions to all the pages that user visits. We try to keep the permissions at a minimum. This pull request will not be merged at it's current state.

MaximRouiller:
Was there any follow-up to this feature? Any way to support the online TFS in other ways?

IndrekV:
Since it seems that there is no one url the TFS server is used, there is no way to add it to the extension. Only possible way to use your custom domain TFS server is to create a custom extension based on Toggl Button and this pull request. Here is a guide on how to add custom domains to the extension 
https://github.com/toggl/toggl-button/wiki/Adding-custom-domains

ReddwarfCro:
Now I removed custom TFS installation And leave only  visualstudio online.

PR #189, opened 2015-01-03
Support for Google Mail

hhermsen:
The button will appear next to the labels and tries to match project on the last label. #131


IndrekV:

PR #207, opened 2015-02-03
Add eventum support

glensc:
add support for [Eventum](https://launchpad.net/eventum) issue tracker.

as this particular issue tracker is not service, but you install it on your own, therefore added to `manifest.json` sample domain under [rfc2606](https://tools.ietf.org/html/rfc2606).

I wonder should this be somehow marked in README? the same problem exists for Trac based installations where you install Trac to your own infrastructure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Add Toggl one-click time tracking to popular web tools.
   - [Google Mail][29]
   - [Taiga][30]
   - [HabitRPG][31]
+  - [Eventum][32]
 
 ## Installing from the Web Store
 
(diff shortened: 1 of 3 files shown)


glensc:
adding screenshot, so you can add it to [where can i find button?](https://github.com/toggl/toggl-button/wiki/Where-can-I-find-the-Button%3F) page after merging!

screenshot

IndrekV:
This can not be merged at it's current state. There's no point in adding the example url to the manifest. Toggl Button does not support adding custom url-s that require global permissions and it is hard to add a correct url if this is mostly a self-hosted service. Maybe forking and creating your own version of Toggl Button is the way to go here.

glensc:
how about adding all but manifest change? then the code is at least in one place.

glensc:
Seems it is no longer possible to install custom built extensions on Windows, MacOSX will follow soon

https://support.google.com/chrome/answer/2811969

What is your suggestion how to still use toggl-button with custom domains?

IndrekV:
Thank you for sharing. We'll look into the possibilities.

glensc:
e0d7bb5befcddf12b7427ddacb40f121e6c9531f

IndrekV:
We have decided not to change the permissions of Toggl Button, for users who need to use custom domains they can use the Developer mode and load extension from folder.

glensc:
did you answer to this question?

> how about adding all but manifest change? then the code is at least in one place.

IndrekV:
Yes I will add everything but the manifest.

IndrekV:
Done.

glensc:
thank you. however i need to make new PR as the code lived in fork for a while

glensc:
f739fc123a3729bc21ad3d7845f829f5b77b1337

btw, you should had added `#207` in your commit message, then github would had autolinked the commit here.

PR #213, opened 2015-02-06
Added Countersoft Gemini

SaarCohen:
Gemini on-demand users can now use toggl


IndrekV:

SaarCohen:
No problem, any ideas when this will be released to the public?

IndrekV:
I'll upload a new version in an hour. It takes maximum an hour to be published in Chrome store. So in two hours or less the new version is up.

PR #219, opened 2015-02-17
Helpscout addition

JoshEpic:
We added support for use with Help Scout. Our team has tested this internally and it is working smoothly.


IndrekV:
Thank you for your contribution. I will release a new version with Help Scout support shortly.

PR #232, opened 2015-03-01
add toggl button to wordpress admin pages

relbraun:


IndrekV:
This pull request changes the permissions so that the extension need permissions to all the pages that the user visits. This is too strong and in it's current state will not be merged to the main extension.

relbraun:
it is so pity because in Wordpress sites all the admin pages look like this and is has not permission more than other sites.
Think about it again.

PR #236, opened 2015-03-08
Enhance project and tag handling in Toodledo integration

kctigerhawk:
Updated the Toodledo integration to handle projects differently.

Toodledo is a very flexible tool and there are a variety of ways that people handle projects.  If you monitor the Toodledo forums you will find that people will use the folder, goal, context or even a tag as the location to indicate the name of the project.

Updated the Toodledo integration to allow the Toggl project name to be entered in any of those locations.

The previous implementation just tacked the folder name onto the description which really clutters up the Toggl interface if you only use a few standard folders with uninteresting names (from a Toggl point of view).

Note: There is another option that people use to represent projects, task/subtask.  Was unable to come up with an integration for that type, but maybe someone else will.

Also, tried to implement ToodleDo tag -> Toggl tag integration, but it doesn't appear to be able to be integrated it yet.  It looks like tags can be specified in the edit form, but not specified through the integration?

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -601,6 +601,14 @@ var TogglButton = {
       }
     } else if (request.type === 'currentEntry') {
       sendResponse({success: TogglButton.$curEntry !== null, currentEntry: TogglButton.$curEntry});
+    } else if (request.type === 'getEditFormData') {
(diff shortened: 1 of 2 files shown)


IndrekV:
This functionality is too tool specific. We have decided to keep all integrations as general as possible. We keep it general so that maintaining 80+ integrations would be reasonable. Thank you for your ideas and work.

PR #242, opened 2015-04-02
Grouping projects by clients in the select dropdown

mihaiaperghis:
This groups projects by clients in the select dropdown instead of
workspaces.

The clients are alphabetically sorted, and only clients with active
projects show up. Projects with no clients show up first under "(No
Client)".

Currently there's no Workspace sorting/grouping available, but this
could be done with a separate select dropdown for Workspaces (if
TogglButton.$user.workspaces.length > 1) situated above the project
select. Based on the Workspace select value, the project select box
would be populated with projects featuring the respective project.wid.

PS: My programming skills are barely existent, so I apologize beforehand if I screwed something up.

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -390,34 +390,50 @@ var TogglButton = {
   fillProjects: function () {
     var html = "<option value='0'>- No Project -</option>",
       projects = TogglButton.$user.projectMap,
-      wsHtml = [],
+      clHtml = {},
       client,
       project,
-      key = null;
-
-    if (TogglButton.$user.workspaces.length > 1) {
-      TogglButton.$user.workspaces.forEach(function (element, index) {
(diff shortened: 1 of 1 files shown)


IndrekV:
Thank you for the contribution. I created a separate solution that keeps the Workspace names and adds clients grouping.

mihaiaperghis:
Thanks!

I tried the changes but the extension crashes. I think it's because wsHtml and clientHtml are arrays, and the workspace/client IDs are very large numbers, thus creating massive arrays with holes.

This is why I used objects instead in my commit, and also used client names instead of IDs in order to use Object.keys().sort() (so the client names appear alphabetically).

IndrekV:
Ok, I'll take another look at it. Thanks for letting me know.

IndrekV:
this commit adds alphabetical sort to clients 85bc4511cdcdf70d696e595b4c6eac47219cd09b

PR #248, opened 2015-04-11
Change the Wunderlist integration to be more subtle and out of the way

iwozzy:
The way the icon was displayed in Wunderlist was sharp. I've adapted it to have the look and feel similar to other icons that are attached to task items in Wunderlist. 

I think this works better because the Toggl button does not distract the user from their task, by blending into the interface style.

image


iwozzy:
Anyone?

IndrekV:

PR #256, opened 2015-05-10
fix sprintly url

christianseel:
simple fix for the url of sprintly


IndrekV:
Thank you for the fix.

PR #257, opened 2015-05-10
added protonet integration

christianseel:
This PR adds a integration for protonet soul (https://protonet.info/)

The toggl button is added to each todo-detail-view next to the other options in the top right.

Integration screenshot:
voila_capture 2015-05-10_05-00-44_nachm

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Add Toggl one-click time tracking to popular web tools.
   - [Flow][37]
   - [Sprintly][38]
   - [Google Calendar][39]
-  - [Protonet][40]
+  - [Protonet SOUL][40]
   
 ## Installing from the Web Store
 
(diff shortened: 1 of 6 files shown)


IndrekV:
Is there a way to test it out with some demo account or something similar.

christianseel:
Hi, actually this is not possible since a protonet server is a private server that companies can host from their offices. If you really need to test it, I could set up a test account on my server, but this one is on a custom domain so it's even more complicated.

IndrekV:
Hi Christian,

As Protonet seems to be quite a custom solution, meaning that it is usually on a private server and on a custom domain, there is no generic url that Toggl Button can use.

christianseel:
Hey, the avarage protonet user is always on `*.protonet.info`, which is the default subdomain for each protonet server.

PR #260, opened 2015-05-14
Automatic Tagging

dbarletta:
- added generic tag recognition logic to the project
- added new option in settings screen to on/off this new feature ("Automatic Tags")
- added specific tag recognition for trello (using trello card labels) [ignores unnamed labels]
- added specific tag recognition for asana (using asana task tags)
- fixed bug: after fetching projects when constructing html to return there was a condition deleting projects from clients with only one project.
  previous conflicting condition was: (wsHtml[key][ckey].match(/\/option/g) || []).length > 1)
  changed to: (wsHtml[key][ckey].match(/\/option/g) || []).length > 0)

### DIFF
--- a/src/html/settings.html
+++ b/src/html/settings.html
@@ -14,6 +14,10 @@
         <label for="show_post_start_popup">Show post-start popup</label>
       </li>
       <li>
+        <input type="checkbox" id="auto-tags">
+        <label for="auto-tags">Automatic tags</label>
+      </li>
+      <li>
         <input type="checkbox" id="websocket">
         <label for="websocket">Live updates (websocket)</label>
       </li>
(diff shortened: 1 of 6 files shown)


syko:
Hi,

Thank you for contributing. I'm terribly sorry for the lack of feedback.

I think we should implement this in a couple more tools before we can release it. Namely Asana, Jira and Todoist.

We can maybe help out with those but the PR needs a rebase first.

bryanph:
Hey @dbarletta,

I realize it has been a while since you submitted this PR (2015 :fearful:). We will close this PR for now as we decide whether to implement this feature or not. If we do, we'll make sure to give you some well deserved credits!

I opened an issue here related to your PR here: #1186

PR #262, opened 2015-05-14
fix toggl-button for new teamweek subdomain

janika:
In Teamweek users are now redirected to app.teamweek.com not to new.teamweek.com.


IndrekV:
Thank you for your superb contribution!

PR #275, opened 2015-05-27
Improved style for Wunderlist.

nibushibu:
before

after


IndrekV:
This was actually just changed with #248 so the toggl Button link would seem more like part of the Wunderlist application. We will currently keep it as it is.

PR #283, opened 2015-06-05
WIP enhance running timer check for the current page

dasilvacontin:
If you prefix or suffix the entry description, it will now correctly
display the timer as running when you navigate/refresh to the same page.


dasilvacontin:
Dang, didn't change the check on websocket update.

dasilvacontin:
I suspect this introduces a bug. I'll fix it and rebase soon.

IndrekV:
The logic has already changed a bit and I think this change is not needed.

dasilvacontin:
Cool, will try to notice if it works now. :)

PR #287, opened 2015-06-09
Fix Teamweek integration

miscer:


IndrekV:
Thank you for your contribution Michal

PR #296, opened 2015-07-06
Wrike support for toggle-button

itainathaniel:
Since my company wants to work with Wrike instead of Trello, I want to contribute to Toggle's button chrome extension


IndrekV:

PR #307, opened 2015-08-03
Apply habitrpg domain name change to habitica.com

Htarlov:
Habitrpg <--> Toggl integration does not work because of domain name change.
Please pull this domain name update.


Htarlov:
I see now that there is already similar pull request from other user. Closed.

PR #308, opened 2015-08-06
Added Assembla integration.

markpavlitski:
This adds integration with Assembla task management https://www.assembla.com/


IndrekV:

PR #323, opened 2015-09-16
Update README.md

glensc:
keep the link clickable


IndrekV:
Thank you for the contribution.

IndrekV:
Actually I'll remove this all together as the Developer mode and loading unpacked extension should continue to work in the future. Just the packaged custom extensions will not be allowed anymore.

glensc:
Do you have link with explanation that backs up your story?

IndrekV:
The article on google talks only about packaged extensions. If unpacked extensions would be blocked it would not be possible to develop extensions so I think this will not happen.

PR #333, opened 2015-09-29
Adds Pivotal Tracker ID to task description.

emilyemorehouse:


dasilvacontin:
Works. It feels logical to have this – if someone changes the task description, you can still match time entry with PT task using this ID. GitHub Toggl Button works this way too. LGTM. :rocket:

Thanks! :)

PR #353, opened 2015-10-24
Added Draftin to this plugin

to0namiSB:
I added a screenshot of where the Toggl button should be at. 
untitled 3


IndrekV:
I've fixed the css. Thank you for your contribution.

PR #354, opened 2015-10-24
Pomodoro mode implementation for toggl-button

Stvad:
I've implemented support for pomodoro-style (https://en.wikipedia.org/wiki/Pomodoro_Technique) regime for toggl-button.
The idea of Pomodoro Technique is to work in limited focused periods of time. So if you enable this mode in the options - you start the tracking work for preset period of time. After that - the tracking would be automatically stopped and you will see the notification (and hear sound signal if you choose so) that will tell you that it's time to take a short break.


src/html/settings.html
```diff
@@ -18,7 +18,7 @@
         <li class="nag-nanny">
           <input type="checkbox" id="nag-nanny">
           <label for="nag-nanny">Remind me to track</label>
-          <div id="nag-nanny-details">
+          <div id="subsettings-details">
```
wmertens: duplicate id with below?

wmertens:
Nice! :+1: (except that duplicate id?)

Stvad:
@wmertens Thanks for pointing that out. There should really have been class-es. I've updated the code.

wmertens:
:+1:

IndrekV:
Thank you very much for your hard work and your contribution to Toggl Button

cassus:
:+1:

PR #358, opened 2015-10-30
Add Discard Time Notification

dasilvacontin:
Changes:

- Renames $timer/$idle to $nanny
- Adds suport for stop date and cb in #stopTimeEntry
- Add Discard Time Notification

Shows a notification to the user when she's been idle for at least 5min while
she was tracking some activity using Toggl.

The notification reads something like:

`You've been idle for 14m 53s while tracking "working on the landing
page".`

And two options are given to the user:

- Discard idle time
- Discard idle time and stop tracking

This notification/feature can be disabled via the new checkbox in the
eettings page, that reads like:

- [ ] Discard idle time notification

### DIFF
--- a/src/html/settings.html
+++ b/src/html/settings.html
@@ -34,6 +34,10 @@
             </p>
           </div>
         </li>
+        <li class="discard-notification">
+          <input type="checkbox" id="discard-notification">
+          <label for="discard-notification">Discard idle time notification</label>
+        </li>
         <li class="pomodoro-mode">
           <input type="checkbox" id="pomodoro-mode">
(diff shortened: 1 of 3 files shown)


dasilvacontin:
Probably some changes are needed for:
- [ ] Ensuring that old settings are still taken into account (now that the local storage key for the nanny feature has been changed)
- [ ] Maybe update the analytics event for the nanny feature could be updated.

Let me know your thoughts. :)

IndrekV:
All is looking good and in master. Changed some names to be similar to idle detection logic used in desktop app.

PR #372, opened 2015-11-16
Make Todoist content script detect project names from different views

imankulov:
Default implementation picks up projects names only from Todoist
project view. This modification extracts names from "project labels"
in filters and labels views.


IndrekV:

PR #381, opened 2015-11-21
Add MantisHub support

vboctor:


vboctor:
How long does it take from the time this PR is accepted till it is updated for users of the add-on via the Chrome store?

vboctor:
@IndrekV just checking if you have noticed this pull request.  Thanks.

IndrekV:
Sorry for not giving any feedback. I have noticed it. We have plans to change our permissions logic and that's why I can't simply merge and release this pull request. Currently every tool addition would need new permissions and the extension will be disabled until the user grants permissions.

The permission logic change would make it possible for users to enable permission to only the domains that they actually use so the update won't disable the extension. We usually do a release every other week or so.

vboctor:
Thanks @IndrekV for the clarification.

I'm wondering if the planned changes will change the way our integration is implemented.  Any reason why not to get this in and treat it like the N other integrations you already have in terms of evolving to the new approach?

Anyway, happy to help in whatever way to get this integrated.

vboctor:
@IndrekV Happy new year.  I've noticed that I got a new update of the toggl plugin.  Unfortunately, this PR wasn't integrated yet following new or old model.  Any updates on next steps?

IndrekV:
I've started reviewing this pull request. In what view should the Start button appear. I haven't managed to see it yet. Could you please share screenshot and description.

IndrekV:
The integration was quite incomplete. The selectors of the start timer button location were broken. I fixed the selectors and improved the styling.

PR #408, opened 2015-12-27
fix button render in new version of gitlab

migaber:
Fix button rendering in new version of gitlab.


IndrekV:

PR #426, opened 2016-01-18
Added Evernote integration

schipa:
toggl-evernote

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Add Toggl one-click time tracking to popular web tools.
   - [Salesforce][50]
   - [Draftin][51]
   - [FogBugz][52]
+  - [Evernote][53]
 
 ## Installing from the Web Store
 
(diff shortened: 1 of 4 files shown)


IndrekV:
Just for information. The pull request will be evaluated and merged after the end of the toggl developer challenge (https://toggl.com/toggl-developer-challenge) on February 14th. Thank you for your contribution.

IndrekV:
I'm not able to make this integration work. `The start timer links are not appearing for me`.

It also seems that you have done too much manual work with Mutation Observer instead of using the observer that is inside common.js

schipa:
I fixed the selectors and the "Start timer" should appear.

The problem with the observer in `common.js` is that the dynamic loading of notes won't remove the `.toggl` class and thus prevents `renderTo` from calling the `renderer` function because `elems` will be empty.

IndrekV:
Merged. I simplified the script and added project matching support. 

There is no need to remove and re-add the timer button. Using functions as `description` and `projectName` values, fetches the correct info when button is pressed.

Thank you for your contribution, sir!

PR #441, opened 2016-01-30
Support for OsTicket: http://osticket.com/

t-oster:
This adds the Button to OsTicket instances.
However, since all OsTicket instances have a custom URL, this has to be enabled manually through manifest.json

### DIFF
--- a/src/scripts/content/osticket.js
+++ b/src/scripts/content/osticket.js
@@ -1,0 +1,22 @@
+/*jslint indent: 2, unparam: true*/
+/*global $: false, document: false, togglbutton: false*/
+
+'use strict';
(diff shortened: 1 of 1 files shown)


IndrekV:
We will change the url logic after the campaign to make it easier for users to add custom url tools so it will work better in the future.

Just for information. The pull request will be evaluated and merged after the end of the toggl developer challenge (https://toggl.com/toggl-developer-challenge) on February 14th. Thank you for your contribution.

IndrekV:
We've decided not to add this integration to Toggl Button. Thank you for participating in the Toggl Button Challenge and for all your work.

PR #442, opened 2016-01-30
Feature: Support for Gogs (https://gogs.io/)

t-oster:
Gogs is similar to Github/Gitlab, but runs on more platforms and is easier to setup.
Since gogs also runs on custom domains, the domain has to be setup manually in manifest.json

### DIFF
--- a/src/scripts/content/gogs.js
+++ b/src/scripts/content/gogs.js
@@ -1,0 +1,21 @@
+/*jslint indent: 2, unparam: true*/
+/*global $: false, document: false, togglbutton: false*/
+
+'use strict';
+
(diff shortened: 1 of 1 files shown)


IndrekV:
We will change the url logic after the campaign to make it easier for users to add custom url tools so it will work better in the future. With the update it is possible to setup custom domains from the settings view.

The pull request will be evaluated and merged after the end of the toggl developer challenge deadline (https://toggl.com/toggl-developer-challenge) on February 14th. 

IndrekV:
Merged with rebase. Thank you for your contribution.

PR #467, opened 2016-02-15
Add evernote support

JustinDFuller:
I added evernote support for the contest. It shows up with the cards that display the title of the note.


IndrekV:
It seems that if I have multiple notes all the start time links are attached to the first note. Clicking the link gets data of the correct note.

IndrekV:
Evernote integration added with pull request #426

Thank you for your contribution and for taking part in the challenge.

PR #475, opened 2016-02-24
Sourcelair support

mariosant:
This PR adds support for @sourcelair IDE. It supports project name and description.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+indent_style = space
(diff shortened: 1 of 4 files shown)


IndrekV:
We use the `{observe:true}` parameter to enable the MutationObserver that detects if dom is changed. I've modified the integration to use the observer.

Also the `.editorconfig` file is obsolete (gitignore rule is missing maybe)

Thank you for your contribution!

PR #485, opened 2016-03-02
Updated teamwork projects projectName to only include text and not to…

matthewdickinson:
… include other tags in the same H1.


IndrekV:
Same issue was solved with 2e4bf1efa2e9c7df1a568f1ac2f85516ac85dba6

PR #486, opened 2016-03-03
Add teamwork link to toggl desc

matthewdickinson:
Added an option to add link to teamwork tasks/tickets into the description for the toggl timers. Also added support for Teamwork Desk.

### DIFF
--- a/src/html/settings.html
+++ b/src/html/settings.html
@@ -8,6 +8,7 @@
     <ul class="tab-links">
       <li class="tab-link active" data-tab="1">General</li>
       <li class="tab-link" data-tab="2">Pomodoro Timer</li>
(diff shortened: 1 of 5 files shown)


eatskolnikov:
@matthewdickinson can you add screenshots of the ui changes?

eatskolnikov:
@matthewdickinson The person who merges PR's will probably request UI screenshots so i'm telling you in advance. Could you add some screenshots of the UI?

IndrekV:
We try to keep the extension as generic as possible and try to avoid one integrations specific setting and preferences. The reason is that we have 73 integrations and keeping them all up to date and having custom features on many of them would make things very complex. I will review this to see if this is something that can be made into a generic setting. Meaning that there is one setting for all integrations and the integrations that support this can use it.

Issue #407, opened 2015-12-21
Start and stop timer when chrome is started and closed


IndrekV:
User feedback
>would it be possible to add into Toggl button option "start automatically”? To be able to start time tracking when chrome starts and (optionally) stop it when Chrome is closed?

PR #490, opened 2016-03-09
Added start and stop automatically settings as requested in issue #407

eatskolnikov:
Issue #407 
Added a "Start Automatically" option in settings.
If this option is checked and after opening the browser, the latest task will be started automatically.

Make sure the setting is checked
screen shot 2016-03-09 at 6 14 04 pm

Start tracking a task:
screen shot 2016-03-09 at 6 13 28 pm

Stop it:
screen shot 2016-03-09 at 6 13 37 pm

Close chrome:
screen shot 2016-03-09 at 6 14 38 pm

Open chrome:
screen shot 2016-03-09 at 6 15 00 pm


eatskolnikov:
The stop automatically setting works like this:
Make sure the setting is checked
screen shot 2016-03-11 at 8 28 43 am

Start tracking:
screen shot 2016-03-11 at 8 29 03 am

Close the browser and open it again, the timer will be stopped
screen shot 2016-03-11 at 8 51 33 am

IndrekV:
I've looked into this pull request and it seems that the stoping part does not work reliably. It works if you close window one by one, but does not work if you close browser with `cmd+Q`. This needs some more looking into before its ready to be merged.

Thank you for you work.

eatskolnikov:
@IndrekV Thanks for the feedback and I'll look into it

eatskolnikov:
@IndrekV sorry for the delay. I have a fun story about this issue. At first I was gonna use Chrome custom commands thing (https://developer.chrome.com/extensions/commands) but then ended up doing something just adding:
```js
window.onbeforeunload = function () {
  openWindowsCount = 0;
  TogglButton.stopTimeEntry(TogglButton.stopTrackingOnBrowserClosed);
};
```
This onbeforeunload function is only executed when chrome is completely closed, as is the case when the cmd+q shortcut is used.

IndrekV:
I had to add some more functionality to make it work when browser started but all is good now. 

PR #492, opened 2016-03-14
do git merge --squash

glensc:
so, i propose that you do `git merge --squash` when merging pull-requests

then you get:
a) 1 commit in your `master` branch
b) the detailed commits can be still be viewed from pull request

downsides that i see:
 * committer may go wrong, but the author should be still reporter
 * can't use github webui to do merges, need to click `cli commands` near merge button

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ If you use a setup, where one of the supported services is on a custom domain yo
 ## Contributing
 Want to contribute? Great! Just fork the project, make your changes and open a [Pull Request][99]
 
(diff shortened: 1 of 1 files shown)


IndrekV:
Yes, squashing is a good idea. Thank you for this contribution.

glensc:
please realize that you have to do the squash-merge, not the contributor :)

IndrekV:
I've decided to revert this change as the squashing should be more of a guideline than a strict rule. If the commit messages make sense and are logical there is no big benefit in squashing things together. New integration pull requests can mostly be squashed.

PR #493, opened 2016-03-14
Update manifest.json

glensc:
fix broken manifest. #190

the file is uppercase TFS.js, but referenced as tfs.js in manifest. fix to match the case otherwise plugin does not load on linux

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -419,7 +419,7 @@
     },
     {
       "matches": ["*://*.visualstudio.com/*"],
-      "js": ["scripts/content/tfs.js"]
+      "js": ["scripts/content/TFS.js"]
     },
     {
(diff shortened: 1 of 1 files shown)


glensc:
@IndrekV ping. current `master` is broken on case sensitive filesystems (linux)

glensc:
you may wish to rename the file lowercase in repo instead. but you can't do the rename easily on case insensitive filesystem (such as OSX hfs or Windows ntfs/vfat). tfs.js and TFS.js are indentical on such system.

i believe (not tested) you can do the rename this way:
```
git mv TFS.js tfs.temp
git mv tfs.temp tfs.js
git commit .
```

but i can recreate this PR with file rename if you wish.

IndrekV:
I've renamed the file.

PR #494, opened 2016-03-14
update eventum button integration. #207

glensc:
update integration as mentioned here:
https://github.com/toggl/toggl-button/pull/207#issuecomment-141038193


glensc:
@IndrekV ping!

Issue #362, opened 2015-11-03
Pomodoro updates


IndrekV:
- [x] Show progress somehow. Like in the popup ad a progressbar or in the extension icon.
- [x] Add more settings. User can define if the timer stops when time ends or just notifies and keeps tracking.
- [x] Restart timer continues the exact same entry that was running last. (should maybe be setting)

pdxrlk:
I agree with these, and I'd add:

- Add support for short (5 min) and ling (15 min) breaks. 

Right now, when I "take a break" the timer stops, and that's it. I need to remember to focus again in 5 minutes. It's much more useful IMO if "take a break" initiates a true break timer (with a "Back to work" button when done).

IndrekV:
Comment from user

>It would be even nicer if once toggl stopped the timer after 25m, the next restart notified you of 5m (or a set break time). I suppose turning off the time on that second stretch could be a hassle for some people, but just a ping saying break time is up would be great. There’s no visual on the actual timer, so it’s easy to go over your break time.

rosshinkley:
@pdxrlk @IndrekV I might even go a step further and say the break time should be customizable and also be tracked.  Breaks ought to be functionally equivalent to a pomodoro, complete with the reminders at the end.  I like the "back to work" button idea for this.  Thoughts?

IndrekV:
I've already started looking into the possibility of adding a break timer that notifies when it's time to start tracking again.

eatskolnikov:
@IndrekV how about this to show the progress, using a badge (https://developer.chrome.com/extensions/browserAction#badge) we could show the remaining time in text and i propose the following rules:

If the time is greater or equal to 100 minutes (is an edge case but bare with me) show 99+m 
![screen shot 2016-04-04 at 10 22 12 pm](https://cloud.githubusercontent.com/assets/1693000/14269606/3bd4e514-fab4-11e5-8a81-49114c313efe.png)
If the time is less than 100 minutes then show the number of minutes remaining like this: 25m 
![screen shot 2016-04-04 at 10 22 55 pm](https://cloud.githubusercontent.com/assets/1693000/14269611/50f047b8-fab4-11e5-9081-d4f245ad2765.png)
If the time is less than a minute start counting the seconds: 39s
![screen shot 2016-04-04 at 10 23 17 pm](https://cloud.githubusercontent.com/assets/1693000/14269616/635ae3a4-fab4-11e5-97ed-e2f1543ade65.png)

[...]

PR #495, opened 2016-03-15
Pomodoro updates - #362

kretel:
Adding extra setting as per request  https://github.com/toggl/toggl-button/issues/362

- Add more settings. User can define if the timer stops when time ends or just notifies and keeps tracking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ If you use a setup, where one of the supported services is on a custom domain yo
 ## Contributing
 Want to contribute? Great! Just fork the project, make your changes and open a [Pull Request][99]
 
-Beware that your commits will be sqashed when merging. This keeps the git log more compact and clear.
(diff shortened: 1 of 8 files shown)


IndrekV:
I think the action buttons in the notification should also be different depending on the setting.

If timer is stopped it should be the same as right now:
- Continue latest
- Start new

If timer is not stopped by pomodoro:
- Stop
- Stop and start new

kretel:
Agreed. Will apply the changes.

eatskolnikov:
@kretel I took the liberty of fixing the conflict and added the buttons @IndrekV mentioned.
Here is the PR https://github.com/toggl/toggl-button/pull/514 (:

kretel:
@eatskolnikov thank you for doing this. 

@IndrekV, do you think you can merge the PR #514 now ? 

Thanks,
Kris

Issue #245, opened 2015-04-07
Automatic timer stop at workday end


IndrekV:
Add a setting where user can input the time that is the end of the day and the timer should definitely be stopped.

IndrekV:
Done with PR #511

PR #511, opened 2016-04-04
Timer can now be stopped at workday end, as requested in #245 

eatskolnikov:
- Added settings option for checkbox to turn on/off the feature
- Added settings option to set the time the workday ends
- Tried to fix issue with settings screen, it overflows when options are expanded
- Added notification when timer stops
- Added button in notification to continue the recently stopped entry

The notification
screen shot 2016-04-04 at 2 34 45 am

Checkbox and input field
screen shot 2016-04-04 at 2 33 04 am


IndrekV:
I've made some changes so that the interval is created only when the setting is enabled.

Good work on this feature.

Issue #477, opened 2016-02-25
Make Pomodoro work also when timer was started elsewhere


IndrekV:
Probably should be a setting. If enabled pomodoro timer is triggered when info comes in about running timer.

gianpaj:
As in for example starting the timer in the Toggl website? I agree with that.

IndrekV:
Starting from anywhere else than Toggl Button. Website, mobile, desktop app, any other tool that connects with toggl API.

PR #512, opened 2016-04-04
Added pomodoro working when starting from elsewhere, as requested in #477

eatskolnikov:
I literally just started the pomodoro timer when the fetchUser method is over if there is currently a time tracking entry. To test it I set the pomodoro time to a minute and used the android app to start the timer.


IndrekV:
Thank you for your contribution. This solution was surprisingly simple.

eatskolnikov:
@IndrekV Indeed, i was also very surprised

PR #514, opened 2016-04-04
Some pomodoro updates taking off from PR #495

eatskolnikov:
I've noticed PR #495 has remained quiet for 20 days, so i took the liberty of finishing it.

Here is the new setting, the checkbox at the bottom of the image
screen shot 2016-04-04 at 12 04 42 pm

This is how the notification looks like if the checkbox is not checked
screen shot 2016-04-04 at 11 50 54 am

### DIFF
--- a/src/html/settings.html
+++ b/src/html/settings.html
@@ -83,7 +83,8 @@
           <li class="pomodoro-mode">
             <input type="checkbox" id="pomodoro-mode">
             <label for="pomodoro-mode">Enable Pomodoro mode</label>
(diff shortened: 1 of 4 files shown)


IndrekV:
Just a note about stoping time entries. If new time entry is created previous running entry will be stopped automatically. No need to explicitly call `stopTimeEntry`

Thank you for your work. 

Have you already contacted our support @eatskolnikov ? We would like to send a swag package your way for your contributions. If you haven't please contact us via support[at]toggl.com and say that the email is for Indrek.

eatskolnikov:
@IndrekV good to know about the stop time entries fact. I contacted support the last time you suggested, tänu kutt (that's supposed to mean "thanks dude" in estonian :p)

PR #522, opened 2016-04-11
Supporting Bitrix24

matheusmatos:
Hello guys,

I tried create a support for Bitrix24. I would that when user click at "Start Execution" on Bitrix24, trigger the Toggl timer.

Can you guys check this code? I made inspired on asana.js.

Thanks a lot

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -39,6 +39,7 @@
     "idle",
     "*://*.toggl.com/*",
     "*://toggl.com/*",
+    "*://*.bitrix24.com/*",
     "*://*.teamweek.com/*",
     "*://*.pivotaltracker.com/*",
     "*://*.github.com/*",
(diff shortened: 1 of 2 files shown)


IndrekV:
I've checked out the pull request and it does not work. I don't think it's a good idea to add the toggl start event to the bitrix24 "Start Execution" button. I think it makes sense to add a separate toggl button start timer link that can be used to start toggl timer. This is how it works in all the other integrations. You can see how the button looks in other integrations on this page https://github.com/toggl/toggl-button/wiki/Where-can-I-find-the-Button%3F

Let me know what you think of this. If you are ok I'll look more into making it work.

jayvdb:
Note I think this has been broken in https://github.com/toggl/toggl-button/issues/899

Issue #501, opened 2016-03-28
Buttons appear multiple times on Wordpress.com


orschiro:
Dear all, 

Can someone reproduce the issue that the Toggl Buttons are shown multiple times?

![nimbus-image-1459155222804](https://cloud.githubusercontent.com/assets/170947/14074952/dcd15184-f4cc-11e5-9d4d-f05b66e1d5e1.png)

Thankfully,

~Robert

IndrekV:
This should not happen as we check that there is not button present before we add a start timer button. I cannot reproduce this. Could you give more details so I can investigate.

orschiro:
Dear Indrek,

Thank you for your response.

I don’t know, however, which additional information to provide you with.

It concerns the posts as well as pages editor of a Blog hosted on
Wordpress.com.

Moreover, I am using the latest version of Toggl Button as well as
Chromium.

Does this help?

Thankfully,

~Robert

On Monday, March 28, 2016, Indrek V notifications@github.com
<http://mailto:notifications@github.com> wrote:

This should not happen as we check that there is not button present before
> we add a start timer button. I cannot reproduce this. Could you give more
> details so I can investigate.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/toggl/toggl-button/issues/501#issuecomment-202374538>
>
​

IndrekV:
It seemed to happen when scrolling the page.

orschiro:
Thanks for rolling out that quick fix!

Warmly,

~Robert

PR #525, opened 2016-04-19
Avoid multiple start timer links in Redmine

chieping:
If an issue has h2 elements, toggl button appears multiple times.

For example:
http://demo.redmine.org/issues/130734

I referred to #501 and applied same solution.
The solution also worked.

Would you please check my pr?

Thank you.

### DIFF
--- a/src/scripts/content/redmine.js
+++ b/src/scripts/content/redmine.js
@@ -8,6 +8,10 @@ togglbutton.render('body.controller-issues.action-show h2:not(.toggl)', {}, func
     titleElem = $('.subject h3'),
     projectElem = $('h1');
 
+  if (!!$('.toggl-button')) {
+    return;
+  }
+
   description = titleElem.innerText;
   if (numElem !== null) {
     description = numElem.innerText + " " + description;


chieping:
FYI I tried another approach what is fixing selector such as adding `:first` or `:first-of-type` before my pr's approach but these did not work.

eatskolnikov:
@chieping Why not just let the other timers appear aligned to their respective h2's? Here is what i mean:
Before:
screen shot 2016-04-20 at 9 16 11 am

After:
screen shot 2016-04-20 at 9 31 01 am


And here is how i think the file should end up looking:

```JS
togglbutton.render('body.controller-issues.action-show h2:not(.toggl)', {}, function (elem) {
  var link, description,
    titleElem = $('.subject h3'),
    projectElem = $('h1');

  description = titleElem.innerText;
  description = elem.innerText + " " + description;

  link = togglbutton.createTimerLink({
    className: 'redmine',
    description: description,
    projectName: projectElem && projectElem.textContent
  });
  elem.appendChild(link);
});
```

IndrekV:
The original pull request solution was the best. There's no point in attaching "Start timer" button to all `h2` links as they are inside the description text.

PR #528, opened 2016-04-20
Showing pomodoro progress with a line, is related to issue #362

eatskolnikov:
Here is another way to show the pomodoro progress using canvas

screen shot 2016-04-20 at 10 03 11 am
screen shot 2016-04-20 at 10 03 19 am
screen shot 2016-04-20 at 10 03 38 am


IndrekV:
Thank you for your work. I'm going to be less active this week so I'll continue with this next week and we'll get the animated pomodoro merged.

IndrekV:
I think we'll go ahead with the circle one as it seems to look better.

Issue #480, opened 2016-02-26
Add support for default project


IndrekV:
If default project is defined then that project will be selected if no project matching is available for the used integration.

eatskolnikov:
I'm confused by the description. What i understand by the title is that the user will be able to setup a default project in settings, making it so when they click the button instead of seeing something like this:
![screen shot 2016-03-07 at 7 52 08 pm](https://cloud.githubusercontent.com/assets/1693000/13587385/5211e376-e49e-11e5-9363-834227553642.png)
they'll see something like:
![screen shot 2016-03-07 at 7 52 31 pm](https://cloud.githubusercontent.com/assets/1693000/13587399/5eb09e74-e49e-11e5-9216-836f1041fef3.png)

Am I wrong?

IndrekV:
Similar feature is present in our desktop app. Default project is selected when there is no project while starting the timer. This happens when user:
- starts timer from the dropdown
- Starts timer from a tool that does not have project matching support or the project name in integration does not match name in toggl.

PR #531, opened 2016-04-22
Default project setting added

eatskolnikov:
As requested in #480 
- Setting shows dropdown with caption in the same format as in desktop app, <project name> . <client name>
- Default project is selected in popup
- Default project is selected in popup for integrations
screen shot 2016-04-22 at 12 54 17 am
screen shot 2016-04-22 at 12 54 30 am


IndrekV:
I made some small style changes, but other than that great work @eatskolnikov.

PR #540, opened 2016-05-03
Adding more Toggl Button areas to eProject.me

topdown:
Adding Toggl Buttons to the Recent Tasks table on the dashboard and also to the full tasks table.
screen shot 2016-05-03 at 12 17 02 pm


topdown:
Something went horribly wrong here with my merge There should only be one file changed not 47.

PR #578, opened 2016-06-14
Add exana.io (Phabricator)

altendky:
The task view still doesn't work quite right but Phacility works better than before and the version of Phabricator running on Exana is now supported as well.


IndrekV:
I made some adjustments for description and project task fetching but most of it worked ok.

PR #589, opened 2016-07-04
Added autocomplete for Project and Tag selection

IndrekV:
Main functionality is in `autocomplete.js`

The dropdowns are currently used in `popup` and `common.js` (the script that injects toggl button to integration page html)

All the dropdown contents are generated in the background script `getEditForm` method.


leafi:
Yo - apologies for taking so long.

Spotted a couple of HTML escaping problems; think it's just literally lack of escaping, at least for tags:
(project name `b<b>wahey!</b>b`, client `c<b>wahey!</b>c`, tag `t<b>wahey!</b>t`:)

screen shot 2016-07-12 at 10 42 38

screen shot 2016-07-12 at 10 42 59

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
...
+};
+
+AutoComplete.prototype.addNew = function (text) {
+  var val = text || this.filter.value,
+    list = this.el.querySelector("." + this.type + "-list"),
+    item = document.createElement("li");
+
+  item.className = this.type + "-item selected-" + this.type;
+  item.setAttribute("title", val);
+  item.innerHTML = val;
```
leafi: Perhaps e.g. `.textContent` instead? (or .innerText if that works)

IndrekV:
I replaced innerHTML with innerText and escaping seems to be working ok now.

leafi:
Getting some issues with the project dropdown:

Master branch, no problem:

button-master-linux

Autocomplete branch on Mac & Linux:
(`undefined` is a notable symptom here.)

button-auto-linux
button-auto-mac

Everything aside from 'No Project' acts like plain text. The text can be selected by dragging the cursor with LMB held down.

Clicking projects that aren't 'No Project' does nothing.

leafi:
Hit me up on Slack if you'd like VNC access or something. I haven't tried this on a fresh new Toggl account, but it happens both with my personal account here and an account I have linked to the Toggl org.

(I'm not doing anything weird with the projects, basically.)

src/scripts/background.js
```diff
@@ -904,14 +919,14 @@ TogglButton = {
...
-        wsHtml[element.id][0] = '<ul class="ws-list" data-wid="' + element.id + '" title="' + element.name + '"><li class="ws-row' + hideWs + '" title="' + element.name.toUpperCase() + '">' + element.name.toUpperCase() + '</li>' +
+        wsHtml[element.id][0] = '<ul class="ws-list" data-wid="' + element.id + '" title="' + escapeHtml(element.name) + '"><li class="ws-row' + hideWs + '" title="' + escapeHtml(element.name.toUpperCase()) + '">' + escapeHtml(element.name.toUpperCase()) + '</li>' +
           '<ul class="client-list" data-cid="0">';
       });
 
       // Add client optgroups
       for (i = 0; i < keys.length; i++) {
         client = clientNames[keys[i]];
-        wsHtml[client.wid][client.name + client.id] = '<ul class="client-list" data-cid="' + client.id + '"><li class="client-row" title="' + client.name + '">' + client.name + '</li>';
+        wsHtml[client.wid][client.name + client.id] = '<ul class="client-list" data-cid="' + client.id + '"><li class="client-row" title="' + client.name + '">' + escapeHtml(client.name) + '</li>';
```
leafi: Missed an `escapeHtml` here (`title="' + client.name + '">'`)

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
+/*jslint indent: 2, unparam: true, plusplus: true*/
+/*global document: false, window: false, XMLHttpRequest: false, chrome: false, btoa: false, localStorage:false */
```
leafi: maybe misunderstanding how this works; but `window`, `XMLHttpRequest`, `chrome`, `btoa` and `localStorage` not used in this file -> only need `/*global document: false */` ...?

src/scripts/background.js
```diff
@@ -910,80 +916,75 @@ TogglButton = {
...
-          if (wsHtml.hasOwnProperty(key)) {
-            Object.keys(wsHtml[key]).sort();
-            for (ckey in wsHtml[key]) {
-              if (wsHtml[key].hasOwnProperty(ckey) && validate(wsHtml[key][ckey])) {
-                html += wsHtml[key][ckey] + "</optgroup>";
-              }
+      // create html
+      for (key in wsHtml) {
+        if (wsHtml.hasOwnProperty(key)) {
+          Object.keys(wsHtml[key]).sort();
```
leafi: Does this actually sort the properties in `wsHtml[key]`? From what I can tell it shouldn't..

Apparently ES6 spec officially formalizes order of iteration on an object (http://stackoverflow.com/questions/5467129/sort-javascript-object-by-key#answer-31102605) but basically it's creation order.

Short proof Object.keys().sort() doesn't seem to work:
```
> o = {"z": 2, "a": 1}
>>> Object {z: 2, a: 1}
> Object.keys(o).sort()
>>> ["a", "z"]
> for (k in o) { console.log(k); }
>>> z
>>> a
```

=> if keys were sorted at time of creation, remove line. else, something like `var sortedKeys = Object.keys(wsHtml[key]).sort(); for (ckey in sortedKeys) { ... `

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
...
+    task;
+
+  if (!!pid) {
+    project = this.el.querySelector("li[data-pid='" + pid + "']");
+    if (id !== 0) {
+      if (!!project) {
+        className = project.querySelector(".project-bullet").className;
+        elem.className = className;
+        result = " - " + project.getAttribute("title");
+        task = project.querySelector("li[data-tid='" + tid + "']");
```
leafi: Some calling code checks if (!!pid), if (!!tid) and if not individually replaces them with 0 -- but some calls to `setProjectBullet` don't. Is it OK to use straight tid here?

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
...
+  var project,
+    className,
+    id = parseInt(pid, 10),
+    elem = el || this.placeholderItem.querySelector(".project-bullet"),
+    result,
+    task;
+
+  if (!!pid) {
+    project = this.el.querySelector("li[data-pid='" + pid + "']");
+    if (id !== 0) {
```
leafi: `id === 0` when `parseInt(pid, 10) == 0`. I'm speculating that when this is true we won't enter `if (!!pid)` b/c 0 is false. So this check can be removed?

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
...
+
+  if (parseInt(id, 10) === 0 || !selected) {
+    return "Add " + type;
+  }
+  return result;
+};
+
+ProjectAutoComplete.prototype.filterSelection = function () {
+  var key,
+    that = this,
```
leafi: No callbacks in this func => `that = this` provides no utility?

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
...
+  item.className = this.type + "-item selected-" + this.type;
+  item.setAttribute("title", val);
+  item.textContent = val;
+
+  list.insertBefore(item, list.querySelector("li:first-child"));
+  this.filter.value = "";
+  this.filterSelection();
+};
+
+AutoComplete.prototype.toggleTaskList = function (elem) {
```
leafi: Only called by ProjectAutoComplete => move to subclass?

src/styles/popup.css
```diff
@@ -398,7 +370,9 @@ hr {
 #toggl-button-edit-form #toggl-button-task-placeholder,
```
leafi: No such thing as `toggl-button-task-placeholder` after this commit

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
...
+
+AutoComplete.prototype.addEvents = function () {
+  var that = this;
+
+  that.placeholderItem.addEventListener('click', function (e) {
+    that.filter.focus();
+  });
+
+  that.filter.addEventListener('focus', function (e) {
+    this.parentNode.classList.add("open");
```
leafi: up to you, but would rather do `that.filter.parentNode` here for clarity

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
...
+  for (i = 0; i < s.length; i += 1) {
+    tag = s[i].textContent;
+    tags.push(tag);
+  }
+  return tags;
+};
+
+TagAutoComplete.prototype.filterSelection = function () {
+  var key,
+    that = this,
```
leafi: `that = this` but no callbacks

src/scripts/background.js
```diff
@@ -882,12 +890,12 @@ TogglButton = {
   },
 
   fillProjects: function () {
-    var html = "<option value='0'>- No Project -</option>",
+    var html = '<p class="project-row" data-pid="0"><span class="project-bullet project-color no-color"></span>No project</p>',
       projects = TogglButton.$user.projectMap,
       clients =  TogglButton.$user.clientMap,
       clientNames = TogglButton.$user.clientNameMap,
+      hideWs = (TogglButton.$user.workspaces.length > 1) ? "" : " hide",
```
leafi: FYI workspaces aren't hidden in the filter list when you've started typing to begin a search

src/scripts/background.js
```diff
@@ -896,9 +904,7 @@ TogglButton = {
       clientName = 0,
       i,
       validate = function (item) {
-        return item.indexOf("</option>") !== -1 &&
-          !!item &&
-          ((item.match(/\/option/g) || []).length > 1 || item.length > 1);
+        return item.indexOf("project-row") !== -1 && !!item;
```
leafi: if `item` is falsey, unless it is exactly `""` then `item.indexOf()` will throw before `!!item` is validated. if `item` is exactly `""`, then there's no need to check if `!!item`.

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
...
+      selected[i].classList.remove("selected-" + this.type);
+    }
+  }
+  if (!!ids) {
+    if (typeof ids === "object") {
+      for (i = 0; i < ids.length; i++) {
+        this.el.querySelector("li[title='" + ids[i] + "']").classList.add("selected-" + this.type);
+      }
+    } else {
+      if (this.type === "project") {
```
leafi: when is a projectautocomplete not of type project?

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,509 @@
...
+        if (row.classList.contains("project-row")) {
+          row.parentNode.classList.add("filter");
+          row.parentNode.parentNode.classList.add("filter");
+        }
+        if (row.classList.contains("client-row")) {
+          row.parentNode.classList.add("filter-match");
+        }
+
+        if (row.classList.contains("task-item")) {
+          row.parentNode.parentNode.classList.add("filter");
```
leafi: In the future, it would be ideal in cases like `.parentNode` to comment with what e.g. row.parentNode.parentNode actually is. (e.g. `// .project-row`)

The combination of multiple `.parentNode` calls and the non-symmetrical classList adding and removing in this function means in practice a reader doesn't have a great chance of understanding exactly what's happening here. Without help, anyway.

leafi:
Once the missed `escapeHtml` check is addressed: LGTM.

src/scripts/autocomplete.js
IndrekV: Agreed.

src/scripts/autocomplete.js
IndrekV: Agreed.

src/scripts/autocomplete.js
IndrekV: Jep, seems like an artefact from the refactoring.

src/scripts/autocomplete.js
IndrekV: Right you are. No need for this `if` there.

src/scripts/autocomplete.js
```diff
@@ -0,0 +1,499 @@
...
+  };
+};
+
+ProjectAutoComplete.prototype.setProjectBullet = function (pid, tid, el) {
+  var project,
+    elem = el || this.placeholderItem.querySelector(".project-bullet"),
+    result,
+    task;
+
+  if (!!pid || pid === "0") {
```
IndrekV: In the server request return if project is not set it is undefined. This is why `!!pid` works. Also added `!!tid` check.

src/scripts/autocomplete.js
IndrekV: Fixed with 66d676082415d79eadab83bfab1bcb814dfd8e73

src/scripts/autocomplete.js
IndrekV: I agree. I added comments about the magic behind `parentNode` chains.

src/scripts/background.js
IndrekV: I removed the obsolete test `!!item` cf4ee3370a7144c2a796052762c3217244df2450

src/scripts/background.js
IndrekV: You are right this was an relict of some old code that did absolutely nothing.

src/styles/popup.css
IndrekV: true dat, all old styles are still there. I'll remove them.

src/scripts/autocomplete.js
IndrekV: yes, these are unneeded. Copy paste mess probably.

src/scripts/autocomplete.js
IndrekV: I added `!!tid` check there

src/scripts/autocomplete.js
IndrekV: Actually in cases the data comes from api return if there is no project selected the project value is undefined. That is the case that `!!pid` handles.

src/scripts/background.js
IndrekV: Jep, I made it so that they are always hidden if there's only one.

IndrekV:
Ok, it seems I've addressed all the issues that came up. Thank you @leafi for such a detailed review.

PR #595, opened 2016-07-08
Added support for new Teamwork Projects UI

ripexz:
We released a new version of the UI that currently runs alongside the old one.
Currently in public BETA, but we already had a few customers highlighting that the button doesn't work so this should sort it. 😄 

There was no good way of getting task details options since we didn't have any specific classes/IDs so I added a class on the menu on our end - if it's not up yet it'll be there next rollout.


IndrekV:
Please let me know how I can test it and I'll review this Pull request.

ripexz:
@IndrekV You should have gotten an invite indrek@zombierun.eu
You can check here https://1468244403142.teamwork.com/projects/268098-Test/tasks if the old version is working and then click **"Try the new version"** in the header and it'll switch to the new UI, you can then verify it works there.

It should show up alongside each task in the list and also in the header buttons list when you click to go into task details.

ripexz:
Ah, wait a while, I found a couple of bugs. Will sort them and reopen.

ripexz:
Okey dokey, should be ready for use - working here on both standard and TKO UIs 😄

IndrekV:
The Promise based solution seemed a little bit too complex, but it seemed to work so I left it not touched.

IndrekV:
This was merged with rebase.

PR #613, opened 2016-07-22
Add billable editing and billable/tags icons to timer

IndrekV:
This PR adds Billable and tags icon to popup running timer. Hovering tag icons shows tags in tooltip.

__popup timer with icons__

screen shot 2016-07-22 at 12 17 29

#### Editing billable
Billable editing is shown if:
- selected project is in pro workspace
- No project is selected and default project is from pro workspace

Actions:
- Selecting project should set billable enabled if project has billable enabled.
- If billable is enabled and selected project is not billable the billable flag should not be updated.
- If project is in free workspace set billable false and hide billable row.

__Integration page__

screen shot 2016-07-22 at 12 17 55

__Popup__

screen shot 2016-07-22 at 12 17 21

### DIFF
--- a/src/html/popup.html
+++ b/src/html/popup.html
@@ -17,6 +17,7 @@
 
       <ul id="menu" class="view">
         <li class="timer">
+          <div class="tag-icon"></div>
           <div class="project-bullet"></div>
           <div class="edit-button"></div>
           <div class="stop-button">Stop</div>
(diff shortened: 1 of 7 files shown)


syko:
Refreshing to read plain js without all the frameworks and libraries for a change.

I found one issue that's not in master so it's either caused by this PR or by the fact that it's behind master: there is an extra submit button on the in-page form:
image
(from github)

Otherwise LGTM

IndrekV:
jep "submit" button issue is fixed in master.

PR #618, opened 2016-07-30
[YouTrack] Fix project name fetch regression

1ed:
Users can have permission to change the project field and in this case the project name is in an a tag. 
This PR fixes that case and it was working before 6c6b3fa3f857fecadac2f2e4e02dba90d1bdd9ab.


IndrekV:
Thank you for your contribution

PR #633, opened 2016-08-17
Detect name for private Github projects

andreoliwa:
Project name detection is working only for public Github projects, and not for private ones.


IndrekV:

PR #658, opened 2016-09-10
add onlyoffice support

jeremiahrichter:
This commit adds support for onlyoffice.com collaboration software on the Project screen and on the task screen, with subtask support

Main project screen
image

Task screen and subtask
image


IndrekV:
@jeremiahrichter Thank you for your contribution. I've also added onlyoffice.eu domain as it is used for European clients. Also add trimming to prevent empty spaces in description text and threw in some style modification to make the Start timer link be vertically in the center.

PR #672, opened 2016-09-27
Meistertask Integration

hennessyevan:
Will grab ProjectName, Tags, Description from modal view of MeisterTask
tasks.


IndrekV:

I made some small adjustments to make it work without problems.

- Replaces `.prepend` with native `insertBefore` function.
- Fixed margins so the button would fit better
- Removed `meistertask` from permissions. Current system allows enabling permissions from settings and there is no need to force this permission to everybody.

Thank you for your hard work.

PR #676, opened 2016-10-03
Added support for Teamwork Desk

matthewdickinson:
Added timer for Teamwork Desk (teamwork.com)

### DIFF
--- a/src/scripts/content/teamwork.js
+++ b/src/scripts/content/teamwork.js
@@ -133,3 +133,53 @@ togglbutton.render('div#Task div.titleHolder ul.options:not(.toggl), .view-heade
     return;
   });
 });
+
+
+// Ticket View Page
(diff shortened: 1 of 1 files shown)


IndrekV:
I think this solution is too complex. There should be no need to use promises and do backoffs. Observing certain changes in dom should be enough. See latest update to trello integration. It detects the change of certain dom elements and then updates timer link.
https://github.com/toggl/toggl-button/commit/b0d3e4c2fcc6b352d4cd1c5335cc916147f12cba

syko:
I agree. The contribution is appreciated but plz take a look at the other integrations and try to make it consistent with those. You can simply observe changes in the DOM which would trigger the rerendering. If possible also use the mutationSelector to keep the rate of rerendering at a minimum.

Closing this PR for now.

hypertd:
I'm using this currently at work. We use teamwork desk for ticketing. Currently the button isn't showing up on tickets. 

Any Advice?

PR #677, opened 2016-10-04
[Taiga] Fix button in detail pages and add in issues list

bameda:


IndrekV:

PR #678, opened 2016-10-05
Master

hennessyevan:
Fixed an issue where toggle button would grab the title from the
textbox and the static div in meisterTask


IndrekV:

PR #683, opened 2016-10-14
Add Clubhouse support

suda:
This PR adds support for [Clubhouse](https://clubhouse.io/) project management tool.


IndrekV:
I removed the Timeout part. It is not needed when the element you observe is the one that renders after the needed elements.

PR #709, opened 2016-11-16
Update to Asana

### LINKED ISSUE #685: Disappearing button in Asana
> My toggl button within asana keeps disappearing the past few days since I've upgraded to the new interface (I'm not sure if this is related or not).
> I cleared all my cache, disabled the app and reauthorised access and when refreshing Asana the button reappeared on one task, but as soon as I navigate to another task it's disappeared again.
> Case #367261

hhermsen:
Fixes #685

### DIFF
--- a/src/scripts/content/asana.js
+++ b/src/scripts/content/asana.js
@@ -3,12 +3,12 @@
 
 'use strict';
 
-togglbutton.render('.details-pane-body:not(.toggl)', {observe: true}, function (elem) {
+togglbutton.render('#right_pane__contents .SingleTaskPane:not(.toggl)', {observe: true}, function (elem) {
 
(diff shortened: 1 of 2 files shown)


IndrekV:
I'm not sure what's it about but when I'm in asana there is no element named `.SingleTaskPane`.

IndrekV:
I'm not sure what's it about but when I'm in asana there is no element named `.SingleTaskPane`.
The old integration worked for me. I have a free account in Asana and it's more than a year old. 

If you validate that the updated version works ok for you then we should just keep both.

hhermsen:
If you create a new project, the layout code is different than before. Keeping both versions sounds good te me.

IndrekV:
Thank you for your contribution, Sir.

PR #721, opened 2016-12-19
Added support for Backlog

tomcat0090:
This pull request adds support for Backlog.

Currently you can start a timer for toggl from issue detail view, here is a screenshot of the location:
backlog_issue


IndrekV:
I'm having trouble testing this integration as the tool seems to be only in Japanese.

tomcat0090:
Sorry for the late reply.
Thank you

PR #749, opened 2017-01-31
Added Checkvist support

crispinb:
This adds a toggl button to tasks on [Checkvist](https://www.checkvist.com) lists.

The lists can get horribly messy with every task having a toggl button, and this may be inappropriate in many cases -- who wants to time the 'carrots' item on their shopping list?  So I'm adding the button  only if one or more tasks in a given  list is tagged with "#toggl".

Buttons are appended to the end of the task text. This doesn't look great, but I found that justifying them (eg. to the RHS) made it hard to associate buttons with tasks, so I chose function over form. Jony Ive I'm not.

Screenshot of simple test list with toggl buttons enabled:

toggl_checkvist

### DIFF
--- a/src/scripts/content/checkvist.js
+++ b/src/scripts/content/checkvist.js
@@ -1,0 +1,24 @@
+/*jslint indent: 2 */
+/*global $: false, document: false, togglbutton: false*/
+'use strict';
+
+var toggleEnabledForCheckvistList = (function () {
+  var togglTag = $('.tag_toggl');
+  return togglTag !== null;
+}());
+
+togglbutton.render('li.nonDivider:not(.toggl)', {observe: true}, function (elem) {
+  if (!toggleEnabledForCheckvistList) {
+    return;
+  }
+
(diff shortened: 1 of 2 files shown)


LulzAugusto:
Hey @crispinb, thanks for contributing. We'll review it within the next few days.

LulzAugusto:
@crispinb the team discussed your tag approach and we came to the conclusion that it can be a little bit counterintuitive. What do you think about showing the button on hover instead? This way we keep the UI clean and it's easier for the user to find the button.

crispinb:
Agreed, that's obviously a more discoverable approach. Thanks. I'll submit another pr when I find a moment to look up how to do it - I don't really know javascript/css being a humble Android / iOS app dev ;)

bryanph:
Hey @crispinb, we decided to clear up old PRs to make way for some big refactoring that is upcoming. If you wish to continue work on this PR please feel free to reopen or add an issue for adding Checkvist support!

PR #751, opened 2017-02-07
Fix remaining 'Chrome' in settings

martin-damien:
Fix a remaining "Chrome" in the settings (visible inside Firefox :p )


tcrammond:
Thanks for the contribution 😄

Issue #770, opened 2017-02-28
"Start timer" not showing on Remember The Milk


vedr4n:
The issue was originally submitted by a person on Twitter
https://twitter.com/DaylightGambler/status/836505957752520704

I ran a few tests with my own Remember The Milk account and can reproduce the problem. The "Start timer" button isn't showing where's it's supposed to. 
![screen shot 2017-02-28 at 10 39 31](https://cloud.githubusercontent.com/assets/19473351/23399890/3ecfefd8-fda2-11e6-897c-a461d226f732.png)

lunaticare:
@vedr4n please vote for the fix to be merged upstream.

leafi:
Fixed in PR #771 by @lunaticare <3

leafi:
Reopening until we deploy master to the app/extension stores/whatever they're called.

lunaticare:
The vendor updated the app, integration is broken again, the fix is ready.

fonorobert:
@leafi do i understand correctly that this is just waiting for deploy to stores?

syko:
I believe this has been fixed by the last PR.

PR #771, opened 2017-03-01
Fixed "Start timer" not showing on Remember The Milk

lunaticare:
Resolves issue #770

The button is now right to the task title as it worked before.

another task - remember the milk 2017-03-01 21-04-23


leafi:
Thanks very much for this!

Expect a new release of Toggl Button including this change within a week. (We've got to get the Asana fix out, and we can piggyback on that release.)

PR #772, opened 2017-03-02
Fix button in the Asana board view

### LINKED ISSUE #759: Button missing in Asana
The toggl button doesn't appear in asana anymore when the task is not maximised (column mode). Also when the task is maximised, you can start the tracking by clicking on the button but, the project or tag can't be selected.
User has tried re-enabling the extension.
Reported with #423756

LulzAugusto:
~~Button is back to the board view, but the board task detail view is hijacking the 'focus' event in some way that prevents our listener to be triggered. This solution leaves the user with a dropdown that at least opens on click, but not by tabbing to the input itself. ¯\_(ツ)_/¯~~

~~ps: I'm open to suggestions towards a solution that goes around whatever this view is doing with the focus event, but I couldn't find one myself. Curious thing is that you can open the console, create an event `new Event('focus')`, remove all the methods `stopImmediatePropagation`, `stopPropagation`, `cancelBubble`, `preventDefault`, then dispatch it and it will work. Buuuut if I do that inside the click listener (`autocomplete.js L33`) it doesn't work.~~

closes #759

### DIFF
--- a/src/scripts/autocomplete.js
+++ b/src/scripts/autocomplete.js
@@ -1,5 +1,5 @@
 /*jslint indent: 2, unparam: true, plusplus: true */
-/*global document: false */
+/*global document: false, window: false */
 "use strict";
 
 var noop = function () { return undefined; };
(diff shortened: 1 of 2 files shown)


syko:
Hmm. Not sure what asana is doing but I did find a way to fix the tabbing too.
Instead of adding the listener to the filter input we can add it to the window with `useCapture=true`

```javascript
  window.addEventListener('focus', function(e) {
    if(e.target === that.filter) {
      that.filter.parentNode.classList.add("open");
      that.listItems = that.el.querySelectorAll(that.item);
      that.updateHeight();
    }
  }, true)
```

LulzAugusto:
@syko amended my last commit with your suggestion.

Issue #764, opened 2017-02-20
Salesforce Lightning isn't working on custom domains


vedr4n:
The title pretty much says it all. "Start timer" button works/displays on the official Salesforce Lightning site, but not on custom domains. 

So far the issue has been reported by one user. Unfortunately, I don't have a way of reproducing the reported bug. 

---
Desk case: 426638

nicktoggl:
It's very late here and I might edit these tomorrow if they are unrelated, but also 

case/429902

vedr4n:
Another affected user in case 429866.

PR #780, opened 2017-03-08
Fix salesforce permission and button placement

### LINKED ISSUE #768: Toggl Button Missing in Salesforce

### Environment details

* Operation system: Mac OSX 10.10.5
* Extension version: 1.1.2

### Expected behavior

Should be a red Toggl button next to Salesforce "My tasks"

### Actual behavior

Button not found anywhere. Looks like Salesforce has updated their cloud environment, possibly that's what broke the integration.

### Additional details

* screenshots or screencapture
screen shot 2017-02-26 at 10 54 36

LulzAugusto:
closes #768

### DIFF
--- a/src/scripts/db.js
+++ b/src/scripts/db.js
@@ -44,7 +44,7 @@ var Db = {
     if (!TogglOrigins[origin]) {
       // Handle cases where subdomain is used (like web.any.do (or sub1.sub2.any.do), we remove web from the beginning)
       origin = origin.split(".");
-      while (origin.length > 2) { origin.shift(); }
+      while (origin.length > 0 && TogglOrigins[origin.join('.')] == null) { origin.shift(); }
       origin = origin.join(".");
       if (!TogglOrigins[origin]) {
         return null;
--- a/src/scripts/db.js
+++ b/src/scripts/db.js
@@ -42,9 +42,9 @@ var Db = {
     }
 
     if (!TogglOrigins[origin]) {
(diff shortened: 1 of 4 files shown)


syko:
Can you address #764 in this PR as well?

It seems that custom domains are in the form `https://<domain>.my.salesforce.com`
image

LulzAugusto:
Sure. I'll take a look 👍

LulzAugusto:
Ready for another review. Now it should work with the old and the new UI, under a subdomain or not.

src/scripts/db.js
```diff
@@ -42,9 +42,9 @@ var Db = {
     }
 
     if (!TogglOrigins[origin]) {
-      // Handle cases where subdomain is used (like web.any.do, we remove web from the beginning)
+      // Handle cases where subdomain is used (like web.any.do (or sub1.sub2.any.do), we remove web from the beginning)
       origin = origin.split(".");
-      origin.shift();
+      while (origin.length > 2) { origin.shift(); }
```
syko: Still not letting you off the hook. I believe this will fail with SLD-enabled domains such as `.co.uk`.
https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains shows there's a hefty number of such TLDs.
The only catch-all solution I can think of is to check `TogglOrigins` at each iteration and hope that we never introduce an origin that is just `com` or `co.uk`.

LulzAugusto:
@syko Crap I forgot about those, but now it should work. (needs rebase)

PR #783, opened 2017-03-12
Fixed button missing on Remember The Milk (RTM)

lunaticare:
Resolves issue #770

### DIFF
--- a/src/scripts/content/rememberthemilk.js
+++ b/src/scripts/content/rememberthemilk.js
@@ -2,14 +2,15 @@
 /*global $: false, document: false, togglbutton: false*/
 'use strict';
 
-togglbutton.render('.b-db-Om:not(.toggl)', {observe: true}, function (elem) {
+togglbutton.render("div[class^='b-db-'][role='listbox'] [role='option']:not(.toggl)", {observe: true}, function (elem) {
   var getDescription,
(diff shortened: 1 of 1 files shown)


lunaticare:
It now works
toggl-button rtm

syko:
(sry, accidental close)

leafi:
Not a fan of us having to repeatedly

Is it our policy to allow other types of selectors? (Surely we've done this before.) If so, how about...

Outermost: `div[class^='b-db-'][role='listbox'] [role='option']:not(.toggl)`
Description: `div[class^='b-db-'][role='listbox'] [role='option'] span[style*='user-select: text']`
Project: err dunno
Where to inject button: $(`the whole description selector`).filter(":parent") (yes, requires jQuery - for the parent part)

leafi:
Could probably lose the 'b-db-' part for now, but only if they never do `role='listbox'` on something else.

lunaticare:
@leafi fixed as you suggested, selector is more robust now.

leafi:
Sorry, running a bit behind. Will look into this properly tomorrow & get it sorted. (I want to review against the batches of selectors used in the other content scripts, mull over the robustness of these selectors, and get something proper sorted for the project selector.)

PR #794, opened 2017-03-27
Todoist: Use name of containing subproject as Toggl project

vandor:
Todoist displays entries of a subproject below the main project's entries. Currently toggl-button just uses the main project's name as the Toggl project. This commit adds the ability to use the subproject name.

Any depth of project nesting is handled. We determine the name of the project, its parent, grandparent, and so on up the project hierarchy. Then we pass an array of these project names to the createTimerLink function so that each may be tried successively until one of them matches a Toggl project. (The matching doesn't happen until we click the Start Timer and it sends a message to the extension's background code, which I've enhanced slightly to accept a list of project names to look for.)

We determine the parent of a Todoist project by inspecting the indentation of their entries in the sidebar menu.

### DIFF
--- a/src/scripts/content/todoist.js
+++ b/src/scripts/content/todoist.js
@@ -41,15 +41,23 @@ function getProjectNameHierarchy(sidebarEle) {
   return [projectName].concat(getProjectNameHierarchy(parentProjectEle));
 }
 
+function projectWasJustCreated(projectId) {
+  return projectId.startsWith('_');
+}
+
 function getSidebarEle(elem) {
   var editorInstance, projectId, sidebarRoot, sidebarColorEle, sidebarEle;
   editorInstance = elem.closest('.project_editor_instance');
   if (editorInstance) {
     projectId = editorInstance.getAttribute('data-project-id');
     sidebarRoot = $('#project_list');
-    sidebarColorEle = $('#project_color_' + projectId, sidebarRoot);
-    if (sidebarColorEle) {
-      sidebarEle = sidebarColorEle.closest('.menu_clickable');
+    if (projectWasJustCreated(projectId)) {
+      sidebarEle = $('.current', sidebarRoot);
+    } else {
+      sidebarColorEle = $('#project_color_' + projectId, sidebarRoot);
+      if (sidebarColorEle) {
+        sidebarEle = sidebarColorEle.closest('.menu_clickable');
+      }
     }
   }
   return sidebarEle;
--- a/src/scripts/content/todoist.js
+++ b/src/scripts/content/todoist.js
@@ -4,9 +4,9 @@
 'use strict';
 
 togglbutton.render('.task_item .content:not(.toggl)', {observe: true}, function (elem) {
-  var link, descFunc, container = $('.text', elem), description,
+  var link, descFunc, container = $('.text', elem),
     getSidebarEle, getProjectNameHierarchy, isTopLevelProject, levelPattern, getParentEle,
-    getProjectNameFromLabel, sidebarEle, projectNames;
+    getProjectNameFromLabel, getProjectNames;
 
   descFunc = function () {
     var clone = container.cloneNode(true),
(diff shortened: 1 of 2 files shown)


src/scripts/content/todoist.js
```diff
@@ -34,25 +35,68 @@ togglbutton.render('.task_item .content:not(.toggl)', {observe: true}, function
...
   };
 
-  projectFunc = function () {
-    var projectElem, projectLabel;
-    projectElem = $('.project_link');
-    if (projectElem) {
-      return projectElem.textContent.trim();
+  getSidebarEle = function (elem) {
+    var editorInstance, projectId, sidebarRoot, sidebarEle;
+    editorInstance = elem.closest('.project_editor_instance');
```
syko: I think it doesn't work for me because I can't find this element appearing anywhere. Which view am I supposed to be in for this to work?

src/scripts/content/todoist.js
```diff
@@ -34,25 +35,68 @@ togglbutton.render('.task_item .content:not(.toggl)', {observe: true}, function
...
+  getParentEle = function(sidebarEle) {
+    var curLevel, parentClass, parentCandidate;
+    curLevel = sidebarEle.className.match(levelPattern)[1],
+    parentClass = 'indent_' + (curLevel - 1);
+
+    parentCandidate = sidebarEle;
+    while (parentCandidate.previousElementSibling) {
+      parentCandidate = parentCandidate.previousElementSibling;
+      if (parentCandidate.classList.contains(parentClass)) {
+        break;
```
syko: Maybe it's easier to just use a selector to fetch that element directly rather than to look for it one by one?

vandor: Normally I would agree. In this loop I'm looking for the closest previous sibling that has a certain class. Which selector would you use in this case? I believe we'd need a previous-sibling selector, which doesn't seem to exist currently.

src/scripts/content/todoist.js
vandor: You were probably looking at the Inbox screen, which I realized I hadn't handled. Items in the Inbox should now have the Toggl button link appearing next to them.

vandor:
To see the functionality I'm adding:
 1. Create two Todoist projects with names that match two of your Toggl projects.
 1. Add a task to each Todoist project.
 1. In Todoist, make one of the projects a sub-project of the other (drag the project name in the sidebar)
 1. Now click the name of parent project in the sidebar to view it afresh. You'll see tasks from both the parent project and sub-project (each with their own project heading).
 1. Click the 'Start timer' link next to a task in the sub-project.
 1. Verify that the Toggl timer entry that just started has the project named after the sub-project rather than the parent project.

src/scripts/content/todoist.js
syko: Ok. But is it necessary to be this strict about it and look for the closest previous sibling? I was thinking just some kind of `$('indent_' + (curLevel - 1))` selector will do.

src/scripts/content/todoist.js
```diff
@@ -3,9 +3,84 @@
...
+  }
+  parentProjectEle = getParentEle(sidebarEle);
+  return [projectName].concat(getProjectNameHierarchy(parentProjectEle));
+}
+
+function projectWasJustCreated(projectId) {
+  return projectId.startsWith('_');
+}
+
+function getSidebarEle(elem) {
```
syko: I think a more accurate name for this would be `getSidebarCurrentEle` or something as it's looking for the current project item not the whole sidebar. Got me confused for a second. (also the var `sidebarEle`)

syko:
Now I got it working :+1:
LGTM except for the small code comments.

src/scripts/content/todoist.js
vandor: I agree, that's clearer. I'll get it changed.

src/scripts/content/todoist.js
vandor: I see what you're saying. It does need to be this specific, though. There can be more than one project at an indentation level, and the only way to get the parent of the particular child project is to find the immediate preceding project at the parent level. Consider this scenario:

* Errands
* Groceries
  * Vegetables
* Movies

If we want to find the parent project of Vegetables, your selector would find Errands, Groceries, and Movies.

PR #796, opened 2017-03-29
Add namespace to CSS styles

### LINKED ISSUE #779: Toggl Button injects non-namespaced CSS styles
### Environment details

* Operation system: OSX
* Extension version: 1.1.2

### Description

Toggl Button injects [these CSS styles](https://github.com/toggl/toggl-button/blob/d4e2e628c53a439c2d61b8bbae38e5a3f7dd3f9d/src/styles/autocomplete.css#L215-L231) without any TB specific prefixes. This overwrites CSS styles in Teamweek which makes milestones look like this:

screen shot 2017-03-06 at 12 50 11

Phenome:
So we don't conflict with other page styles

I couldn't find any refrence to `color-x` classes. Probably safe to remove. I think it should've been removed here:
https://github.com/toggl/toggl-button/commit/f46081443e82ee87713b72b85b12987907693d4e
Thoughts?

Closes #779

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -888,7 +888,7 @@ TogglButton = {
   },
 
   fillProjects: function () {
-    var html = '<p class="project-row" data-pid="0"><span class="tb-project-bullet project-color no-color"></span>No project</p>',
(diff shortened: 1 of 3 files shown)


Phenome:
As per discussed, probably some other extension interfering with Toggl Button.
Seems to be working.

I also removed the unused CSS classes. Made a reference to that on the commit.

Issue #797, opened 2017-03-31
Add support for Khan Academy


langri-sha:
Feature request to add support for tracking time on [Khan Academy](http://khanacademy.org/).

Will definitely help those who want to track time whilst watching videos or practicing.

langri-sha:
This one was resolved. Thx @Phenome!

PR #798, opened 2017-03-31
added khanacademy

hkalia:
Added the basics, however, I can't get the button to appear. Would appreciate any help. #797


syko:
Hmm, where are you trying to make the button appear exactly?

hkalia:
Anywhere on the page is fine really, although in the top right would be nice, right by the name of the current user logged in.

Phenome:
I saw the issue and started coding the integration before realizing @hkalia had already issued a PR :$

I replaced with my code, which adds the button to the top left.

Let me know what you think.

src/scripts/content/khanacademy.js
```diff
@@ -2,16 +2,28 @@
 /*global $: false, document: false, togglbutton: false*/
 'use strict';
 
-togglbutton.render('#task:not(.toggl)', {observe: true}, function (elem) {
-  var link,
-    description = $('.title', elem).textContent,
-    project = $('.project', elem).textContent;
+togglbutton.render('#tutorial-content [data-reactid="5"]:not(.toggl)', {}, function (elem) {
```
Phenome: I'm not really fond of using this `data-reactid` on the selector as I think this might change someday.
But they're using custom class names, which would be even less reliable.

hkalia:
This is great!

hkalia:
Sry, I clicked the wrong button. Phenome, perhaps create a new pull request with only one commit to make the merge cleaner?

LulzAugusto:
See #829

PR #847, opened 2017-07-04
Remember last project per service/url

### LINKED ISSUE #210: Connect urls and selected projects
It would be really useful if this extension would remember which project was selected last on a specific host/URL. Thus e.g. if I start timers on Github for the project foo/bar and select the project "Bar" once, all future timer starts will have that project preselected. This would allow for starting timers with a single click if you ignore the dialog.

Phenome:
## What does this PR do?
This adds a new option to the settings page:
Remember last project per service/url

## How to test it
Try the option out. It should:
* Remember the last project used for said service (eg. github) or url
* If it's the same as default project from settings, use that one
* When changing the option, should wipe the mapping

Closes #210

### DIFF
--- a/src/html/settings.html
+++ b/src/html/settings.html
@@ -61,6 +61,15 @@
               <div class="description">The default project will be selected when timer is started with no matched project</div>
             </li>
             <li>
+              <label for="remember-project-per">Remember project per</label>
+              <select id="remember-project-per">
+                <option value="false">Use global default</option>
+                <option value="service">Service</option>
+                <option value="url">URL</option>
+              </select>
+              <div class="description">This will remember the last project used on each service/url separately. Overrides default project.</div>
+            </li>
+            <li>
               <input type="checkbox" id="show_right_click_button">
               <label for="show_right_click_button">Show "Start timer" item in right click menu</label>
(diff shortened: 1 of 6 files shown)


src/scripts/background.js
```diff
@@ -286,7 +286,13 @@ var TogglButton = {
   },
 
   updateCurrentEntry: function (data) {
-    var entry = data.data;
+    var entry = data.data,
+      defaultProject = Db.getDefaultProject(),
+      rememberProjectPer = Db.get("rememberProjectPer");
+    if (rememberProjectPer) {
+      Db.setDefaultProject(entry.pid === defaultProject ? null : entry.pid,
```
syko: This bit could use a comment. I went back and forth 6 times trying to figure out how this line works and doesn't mess up the global default project.

src/html/settings.html
```diff
@@ -61,6 +61,15 @@
               <div class="description">The default project will be selected when timer is started with no matched project</div>
             </li>
             <li>
+              <label for="remember-project-per">Remember project per: </label>
```
syko: Minor thing but lose the `:`

Phenome:
Comments adressed:
* removed `:`
* Added comments
* Now tracking only when any default project is set. (yeah, I thought it was a bit odd to be tracking `pids`, but didn't want to change it.)

PR #848, opened 2017-07-04
Adjust settings dialog to prevent overflow

### LINKED ISSUE #845: Settings dialog layout break if screen scaling enabled
### Environment details

* Operation system: Windows 10 64bit; Chrome 59.
* Extension version: 1.2.0

### Expected behavior
image

### Actual behavior
image

* Have a high resolution screen
* Open the settings dialog

Phenome:
## What does this PR do
Adjust settings dialog to prevent overflow, specially on screens with display scaling

Apparently Chrome is misbehaving when display scaling is turned on. We define our settings dialog's width and height, and Chrome makes the dialog window bigger. Sadly, the content does not scale up.

I tried some CSS to make Chrome respect the dimensions, but couldn't. At least there's no more overflow.

## How to test
* Have high resolution screen, change display scaling
image
* If you also have this feature on your OS (Mac, Linux), please let me know how that looks for you
* See issue, before and after.
* Check specially if there are no uneeded scroll bars, and if the scroll bars are not clipped.

Kinda closes #845

### DIFF
--- a/src/styles/settings.css
+++ b/src/styles/settings.css
@@ -147,6 +147,7 @@ header {
   right: 0;
   background: #ffffff;
   z-index: 100;
+  box-shadow: inset 0px -1px 1px -1px #bbb;
 }
 
 header ul {
@@ -170,8 +171,9 @@ header .active, header li:hover {
 .container .tabs {
   position: absolute;
   bottom: 42px;
-  top: 35px;
+  top: 0;
   left: 0px;
+  padding-top: 15px;
   width: 2600px;
   overflow: hidden;
   transition:left 300ms ease-in-out;
@@ -191,7 +193,7 @@ header .active, header li:hover {
 
 .container .tab {
   float: left;
-  height: 323px;
+  height: 343px;
   overflow-x: hidden;
   overflow-y: auto;
   width: 800px;
--- a/src/styles/settings.css
+++ b/src/styles/settings.css
@@ -21,14 +21,23 @@ input[type="number"] {
   width: 50px;
 }
 
+html,
+body,
 .container {
+  height: 400px;
+  width: 800px;
+  max-height: 400px;
+  max-width: 800px;
+}
+
+.container {
+  position: relative;
+  height: 400px;
   width: 800px;
   margin: 20px 15px 0px;
   padding: 0;
   color: #505050!important;
   float: left;
-  min-height: 380px;
-  max-height: 400px;
   overflow: hidden;
 }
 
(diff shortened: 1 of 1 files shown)


syko:
The only problem that I can see on a mac is the double scrollbar, which doesn't go away in this branch either so this PR doesn't seem to be doing anything for me :P
image

Phenome:
Darn it... I need help with this, then :|
If someone on a Mac can also make a commit?
It's really tough to debug, because if you open the `settings.html` on a tab, the CSS looks fine, so it should work. Then you open the settings dialog and it's crap. So you have to test it out :|
We have to work around this Chrome thing.

It really makes a difference in Windows, though... from this:
image

To this:
image

Not pretty. Really wish we could resize the dialog. But at least we don't have two option pages showing up.

I was also having a horizontal scrollbar appear
image
 and this also removes it.


Not really sure how to get rid of that. Perhaps a `overflow:hidden` on the body could work there, since from the screenshot there doesn't seem to be a huge vertical overflow, so maybe.
But... Someone on a Mac need to test it out, sadly :|

Phenome:
Well... From the screenshot, it just seems "scrolled down", probably not it, but worth pointing out.

The real problem is when you enable screen scaling. If I use the whole 4k resolution of my screen, everything is tiny, but nothing is broken. Trouble is with screen scaling turned on. Is that the case?

LulzAugusto:
Yep, tried both with scaling on & off. And what I was trying to show was the content getting clipped before it reaches the navigation at the top, it doesn't happen on master.

Phenome:
If you don't even get the double scrollbar on master, I'll get back to the drawing board... Just tried again on Chrome  60.0.3112.78 on Windows, with display scaling, and got this:
image
This commits really does make it go away, and doesn't break on non-display-scaled screens.

It really seems like a Chrome issue, but couldn't find any discussions about it :|

LulzAugusto:
As an alternative, maybe we should consider using conditional CSS rules for Windows only then.

dooart:
Just pushed a commit to fix the changelog text position on Mac:
image

And another to improve the vertical content scrolling, because it was clipping a little below the header (also adds a border below the header):
image

@Phenome: can you please check if this doesn't break things again on Windows? If everything looks alright: 🚀

Phenome:
This is looking way better today. As assumed, the problem would just "go away" some time when Chrome updated this... It really seemed like a Chrome problem. Giving out wrong dimensions on a display scaled situation.
Report:
* Everything looking alright :D Does not break on Windows
* Looks even better than before. Thx Chrome

PR #854, opened 2017-07-18
Fix asana integration

### LINKED ISSUE #842: Button doesn't show up in personal task list in Asana
In Asana the button only shows up correctly when viewing a task under a project. Opening a task from the "My tasks" list, the button isn't there.

Console shows the following:

screen shot 2017-06-28 at 18 10 45
reported in case 495797

LulzAugusto:
A selector got updated causing the integration to break.

closes #842

### DIFF
--- a/src/scripts/content/asana.js
+++ b/src/scripts/content/asana.js
@@ -61,7 +61,7 @@ togglbutton.render('.BoardCard.BoardColumnCardsContainer-item:not(.toggl)', {obs
   var link,
     container = $('.BoardCardMetadata-left', elem),
     description = $('.BoardCard-name', elem).textContent,
-    project = $('#project_pane_header_name').textContent;
+    project = $('.SidebarItemRow.is-selected').textContent;
 
   link = togglbutton.createTimerLink({
     className: 'asana-board',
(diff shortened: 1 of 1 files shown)


LulzAugusto:
Amended the commit in order to address the issue @syko pointed out.

src/scripts/content/asana.js
```diff
@@ -61,7 +61,7 @@ togglbutton.render('.BoardCard.BoardColumnCardsContainer-item:not(.toggl)', {obs
   var link,
     container = $('.BoardCardMetadata-left', elem),
     description = $('.BoardCard-name', elem).textContent,
-    project = $('#project_pane_header_name').textContent;
+    project = $('.DeprecatedNavigationLink.SidebarItemRow.is-selected').textContent;
```
Phenome: I think we maybe should drop the `DeprecatedNavigationLink` class. `SidebarItemRow` seems enough (tried the querySelector... both returned the same number.
I say this because `DeprecatedNavigationLink` seems like an obvious candidate to be removed soon. Probably will save us some time in the future.

syko: @LulzAugusto I think this is a good point ☝️

PR #865, opened 2017-08-12
Added duration field to extension popup edit view

IndrekV:
The thing I've missed the most. Being able to edit the duration to reflect the correct time I started working. Actually I just miss working on Toggl Button.

This PR adds duration edit field into the Extension icon popup Edit view.

Tested both in Chrome and FF and it seemed to be ok.

**Chrome**
screen shot 2017-08-12 at 23 52 35

**Firefox**
screen shot 2017-08-12 at 23 44 04

Some comments about the code.

- The duration field updates every second until user focuses the field. `PopUp.durationChanged` flag is used to determine if field should be updated or not.

- Firefox needed to validate the input in the time field as it allows to type in non numeric characters. That's why `isNumeric` function was added to PopUp.

- `(arr[2] || 0)` is used because in cases like `03:00:00` the value for some reason is `"03:00"` and this breaks the calculation as `arr[2]` is undefined.

- The duration edit is only visible in the popup and not in the edit form shown on the page

Happy Reviewing!

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -46,17 +46,20 @@ var TogglButton = {
       '<form autocomplete="off">' +
       '<a class="toggl-button {service} active" href="#">Stop timer</a>' +
       '<a id="toggl-button-hide">&times;</a>' +
-      '<div class="toggl-button-row">' +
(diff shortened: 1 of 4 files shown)


IndrekV:
Adding step attribute seems to have worked.

Issue #874, opened 2017-09-01
Toggl Button no longer working with Zube


kadrihvl:
A recent change we made to our cards broke Toggl. Would it be possible to update Toggl to match our new card layout?
CAse #542059

xenvox:
+1
Our team massively use Zube, this issue stops us from using Toggl...

syko:
Fixed by #875

PR #875, opened 2017-09-01
Add Rollbar integration & Fix toggl button not appearing issue on Zube & VisualStudio

kukiron:
PR for integrating [Rollbar](https://rollbar.com), a real-time error tracking & debugging tool for developers. It's free & really easy to set up for any project. 
- The name of the error/issue is the Timer Description & the Rollbar project name is Timer project.

Individual error tracking page - 

rollbar_error-tracking page

 
Passed `make lint`. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Add Toggl one-click time tracking to popular web tools.
   - [Redmine][18]
   - [Remember The Milk][71]
   - [Rindle][83]
+  - [Rollbar][110]
   - [Salesforce][50]
   - [SherpaDesk][86]
   - [Sifter][16]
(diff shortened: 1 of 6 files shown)


kukiron:
Hi, 

The second commit fixes the Toggl button disappearing issues on Zube #874 & VisualStudio (VSTS) #876.

- In [Zube](https://zube.io), the button appears on individual Card, Ticket & Epic page views. 
- In [VisualStudio](https://www.visualstudio.com) (VSTS), the button appears on all WorkItems pages.

Let me know if there's anything needs to be changed.

Thanks.

kukiron:
Hi,

Sorry for taking long to get back to this. I have removed the Toggl button from "Queries" page. The reason is --

1. On Queries page, the DOM doesn't re-render after every new selection. This stops updating the description of the work (remains the same as the previous one of the same work category).
2. It's possible to make the button appear (as in the image in your comment) & fix the styling issue but when the Queries page has lots of similar types of work items (e.g - multiple Epic, Feature, Issue or Task items), the descriptions are not getting updated. Also, the Test detail view doesn't show the Toggl button.

This requires manipulating the DOM more than the scope of injecting Toggl button onto the page. It's important not to provide wrong details of user activity. So, the button appears on details view selected from "Work Items" & "Backlogs" pages but not from "Queries" page. 

You can look into it further. Will be open to any suggestion.

syko:
Thanks!

I haven't looked into it but I believe we should be able to use the mutation observer to listen to specific DOM changes that should trigger an update to the button. `togglbutton.render` accepts a selector as the last parameter for specifying which mutations we're interested in.

However, I want to get these fixes into the next release so I'm merging this and opening up a new issue for adding it to the Queries page.

:+1:

PR #883, opened 2017-09-14
added PagerDuty integration

ivancamilov:
This adds a Toggl button inside the Incidents in [PagerDuty](https://www.pagerduty.com/).

The button is added to the right side column with the rest of the actions available for an incident. I attach a screenshot showing the placement of the button.

pagerduty


ivancamilov:
That was quick, thanks! :)

Phenome:
Still going to take a little more time until we bundle this and some other commits into a new release.
Hang in there :)

PR #900, opened 2017-10-04
WIP: Fix Bitrix24 support

jayvdb:
This only works on the individual task page, whose URL is
obtained using "Copy task link to the Clipboard".

### DIFF
--- a/src/scripts/content/bitrix24.js
+++ b/src/scripts/content/bitrix24.js
@@ -3,22 +3,15 @@
 
 'use strict';
 
-togglbutton.render('#task-view-buttons.start:not(.toggl)', {observe: true}, function (elem) {
+togglbutton.render('#bx-component-scope-bitrix_tasks_widget_buttonstask_1:not(.toggl)', {observe: true}, function (elem) {
 
   var link, descFunc,
-    description = $('.task-detail-header-title'),
+    description = $('.pagetitle').textContent,
     project = $('.task-group-field-label').textContent;
-
-  descFunc = function () {
-    if (!!description) {
-      return description.textContent;
-    }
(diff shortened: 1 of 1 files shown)


jayvdb:
This is not a complete fix, as it doesnt work on tasks that popup over the task list.

On my development version, I can see the toggl button appear on task pages, the hoverover is the title, but it doesnt put the title into Toggl timer description.

jayvdb:
Probably also should keep the old code, and add the new, like #890.
This is because Bitrix24 is self-hosted as well as hosted, and the dont provide free updates after 1 year, so some old customers may be happily using the previous toggl with their old self-hosted Bitrix24.

starfruitsolutions:
I probably should have looked before doing a duplicate request for this. I didn't notice at first, but I'm having the same issue with tasks accessed without reloading the page. I used '.task-view-buttonset:not(.toggl)' as the selector rather than the ID, in case there was a different ID for different views, but it's failing to update with changes. 

I would have thought using "observe:true" would cover the ajax loading. I tried putting the render function in an interval, to see if it would "try again", but it doesn't even enter the callback. I suspect the change that's needed is in the render function itself. I'm wondering if this is an issue elsewhere.

starfruitsolutions:
oooh, it's in an Iframe. That's the problem.

starfruitsolutions:
I got it all working, but I'll need to clean it up before doing a pull request. You don't have access to the styles from inside the iframe, so I have the .toggl-button styles hardcoded atm.

starfruitsolutions:
Here's the pull request: https://github.com/toggl/toggl-button/pull/936

PR #908, opened 2017-10-13
Adds Airtable support

der-On:
This PR adds support for airtable.com. As it is almost impossible to guess tags or projects from an airtable dataset the integration only uses the datasets identifier as the description.

toggl-button_airtable

### DIFF
--- a/src/scripts/content/airtable.js
+++ b/src/scripts/content/airtable.js
@@ -1,0 +1,22 @@
+/*jslint indent: 2 */
+/*global $: false, document: false, togglbutton: false*/
+
+'use strict';
+
+togglbutton.render('.detailViewWithActivityFeedBase .dialog > .header > .flex-auto:not(.toggl)', {observe: true}, function (elem) {
+
+  var link, descFunc,
+    container = $('.justify-center > .relative', elem),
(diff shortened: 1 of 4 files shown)


der-On:
@syko any chance of getting this merged?

der-On:
I've now resolved the conflicts and would really like some feedback on this PR. Are there any things left to be done to get this merged?

src/scripts/content/airtable.js
```diff
@@ -0,0 +1,22 @@
+/*jslint indent: 2 */
+/*global $: false, document: false, togglbutton: false*/
+
+'use strict';
+
+togglbutton.render('.detailViewWithActivityFeedBase .dialog > .header > .flex-auto:not(.toggl)', {observe: true}, function (elem) {
+
+  var link, descFunc,
+    container = $('.justify-center > .relative', elem),
+    description = $('.truncate.flex-auto.line-height-3', elem);
```
Phenome: Sadly those elements don't have more deterministic class names, so I'd go with perhaps `:nth-child` selectors. But probably not that much more reliable anyway.
These work for current interface layout, so this is good to go :)

der-On:
@Phenome thank you!

PR #923, opened 2017-11-04
Added support for the new Google Calendar Modern UI

### LINKED ISSUE #919: Injection in new Google Calendar is not supported
Google just launched a reworked Calendar under: https://calendar.google.com/calendar/r/week
(could be that this is currently for Google Business users only available).

See also: https://support.google.com/calendar/answer/7541906?hl=en

The Toggl Button injection in the new GCalendar is not yet working.
Hope you can add this asap :) Thanks

dr-BEat:
This PR adds support for the new Google Calendar Modern UI and resolves #919.
The Toggl Button is placed in the popup after clicking on an event, next to the delete, menu and close buttons.
image

### DIFF
--- a/src/scripts/content/google-calendar.js
+++ b/src/scripts/content/google-calendar.js
@@ -58,3 +58,28 @@ var observer = new MutationObserver(function (mutations) {
 });
 
 observer.observe($('body'), { childList: true, subtree: true });
+
+//Google Calendar Modern
+
+function insertButtonModern(bubblecontent, description) {
+  var link = togglbutton.createTimerLink({
+    className: 'google-calendar-modern',
+    description: description
+  });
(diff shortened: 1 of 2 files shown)


Phenome:
Also tested this code with classic, and it's still working 👍.

There's one one other issue, though. If I click on another event without closing the popup first, then the Toggl button vanishes :|

Phenome:
Would suggest removing the `:not(.toggl)` from the selector, and adding a check for the `.toggl-button`  presence before running the render code.
```
if ($('.toggl-button', elem)) {
  return;
}
```

dr-BEat:
Thanks for the comment! I fixed the linting errors and updated the code to use @Phenome changes.

PR #924, opened 2017-11-06
Add custom style for YouTrack agile view

Phenome:
Using `float: right` was causing Toggl button to drop to lower line, cropping it.

Added specific rule for card in agile board, so it does not float.
image

Please make sure this doesn't break elsewhere in Youtrack.


langri-sha:
Bugs are gone!

LGTM :rocket:!

PR #931, opened 2017-11-20
Change duration input type to text

### LINKED ISSUE #879: Chrome showing new timers starting at 12:00:00 am

Chrome using details from a user's OS locale/settings. Both occurrences appear to be on OSX but isn't happening for every user

* Operation system: OSX
* Extension version: 1.5.0

### Expected behavior
The timer should be starting at 00:00:00
### Actual behavior
The timer is starting at 
Unable to determine what is causing the issue consistently. Appears to be happening randomly to different users.
### Additional details

* screenshots or screencapture
image
image

case/546968

Phenome:
Prevents incorrectly displaying duration on 12h locale.
Also uses `pattern` to validate input.

Can't test it because no MacOS, but I'd be impressed if it still displays "AM" with `type="text"`.
Anyway, please test... And validate the validator regex :)

Closes #879

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -47,7 +47,7 @@ var TogglButton = {
       '<a class="toggl-button {service} active" href="javascript:void(0)">Stop timer</a>' +
       '<a id="toggl-button-hide">&times;</a>' +
       '<div class="toggl-button-row" id="toggl-button-duration-row">' +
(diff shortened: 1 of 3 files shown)


LulzAugusto:
Tha validator works only if you press "enter", if you click "done" it accepts the value. Maybe you have to trigger validation on click as well? http://take.ms/vDvsd

Phenome:
Finally gave this another go:
* Now validates form before trying to send 
❤️ css validity selectors
* CSS styles for invalid inputs
* CSS style for button with invalid form
Only needed on the popup, as there is no duration input for the embedded form.

Issue #899, opened 2017-10-04
Bitrix24 : Integration broken


jayvdb:
Please follow this guide when creating a new issue. This speeds up the process of replicating the issue and finding a solution.

### Environment details

* Operation system: Linux / Chrome
* Extension version: 1.7.0

(Also tried on other OS/browser)

### Expected behavior

Starting a timer after loading a task from the task list should start the timer for that task.

### Actual behavior

No timer button appears, and
using the context menu "Start timer" creates a task named "My Tasks"

### Steps to reproduce the behavior

1. Go to tasks, list view
2. open a task, which loads a popup window
3. no timer button appears on the task
4. Right click, select "Start timer"
5. The timer starts with task name "My Tasks".

jayvdb:
I've also tested going to the unique URL for the task, and the timer doesnt appear there either.

jayvdb:
I believe they have changed things so that
- the UI region for the button is now class `task-view-buttonset` , 
- the title is `$('.pagetitle')`

Not changed:
- the optional project is class `task-group-field-label`

maritoggl:
User reported the Russian version of the portal being broken (bitrix24.ru). 

[conversations/19689699578](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2100822/conversations/19689699578)

PR #936, opened 2017-12-05
Fix Bitrix24 Integration

starfruitsolutions:
The layout for Bitrix24 has changed since the last commit. This fix leaves legacy support for those who have self-hosted bitrix24 installations past the update support window. In addition, there is a task popover function, which displays in an iframe. The contents of the iframe aren't accessible by the render method, so I added that functionality in the callback. The styles for ".toggl-button" needed to be inserted into the iframe, so I hardcoded them. Not ideal, but it works.


dooart:
I'm afraid we cannot merge this as it is. Hardcoding a big chunk of existing CSS code isn't, as you said, an ideal solution to this problem.

There's an issue now about this: https://github.com/toggl/toggl-button/issues/970. We'll revisit the Bitrix24 integration as soon as we get that done.

bryanph:
Hey @jayvdb , we decided to clear up old PRs to make way for some big refactoring that is upcoming. If you wish to continue work on this PR please feel free to reopen or continue discussion on #899.

Issue #898, opened 2017-10-04
Habitica Integration Gone


TheSoloJake:
Based on Desk cases 561810 and 559363, the Habitica integration with the Toggl Button is no longer available. Communication with Habitica Support has verified that a recent website change has eliminated the current integrations.

Alys:
I'm one of Habitica's moderators. I'm commenting here to say that if anyone has questions about Habitica's website, you can post here and I'll see it and try to help, although fixing this issue is probably (I hope!) a fairly simple matter of seeing what new HTML from the website needs to be put into the integration.

If there's nothing in our website's HTML that the integration can use, we might be willing to consider adjusting the website slightly. I can't promise that because I'm not staff but I'd be happy to take a request for it back to the staff for them to consider.

8Mobius8:
@Alys there is a slight alignment problem, with my solution in #949. Do you have any clever ideas to make it look better? If need be...

<img width="452" alt="screen shot 2018-01-02 at 23 51 26" src="https://user-images.githubusercontent.com/2933379/34509691-06212ff0-f01b-11e7-9e60-a5cb0c841223.png">

Alys:
@8Mobius8 Thanks so much for working on this! Personally I think that looks pretty good as it is, and I'm not actually strong on the new website's CSS code, so I've asked Habitica's staff if they'd like to comment on this.

8Mobius8:
Updated #949 to include alignment fix using `transform: scale(0.85)`. Hopefully that is okay, result looks like this:

<img width="373" alt="screen shot 2018-01-04 at 21 10 53" src="https://user-images.githubusercontent.com/2933379/34592787-91a5d4ca-f194-11e7-9454-45784a533e0b.png">
<img width="417" alt="screen shot 2018-01-04 at 21 10 42" src="https://user-images.githubusercontent.com/2933379/34592790-93bc334e-f194-11e7-8793-c493a1839a21.png">

tcrammond:
Looks like this is resolved, closing issue 🚀

PR #949, opened 2018-01-03
Fix Habitica Integration

8Mobius8:
* Updates CSS selector for zeh magical ⏺ for `habitica.com`.
#898 
screen shot 2018-01-03 at 00 07 33

The alignment with the other icons bothers me a bit, but its still functional. Would like help on that, if knows a fix.

### DIFF
--- a/src/scripts/content/habitica.js
+++ b/src/scripts/content/habitica.js
@@ -4,11 +4,11 @@
 (function () {
   'use strict';
 
-  togglbutton.render('li.task.daily:not(.toggl), li.task.habit:not(.toggl), li.task.todo:not(.toggl)',
+  togglbutton.render('.daily .tasks-list .task:not(.toggl), .habit .tasks-list .task:not(.toggl), .todo .tasks-list .task:not(.toggl)',
                      {observe: true}, function (elem) {
       var link,
-        text = $('.task-text>markdown', elem).textContent.trim(),
-        container = $('.task-meta-controls', elem);
+        text = $('.task-title', elem).textContent.trim(),
+        container = $('.icons-right', elem);
 
       link = togglbutton.createTimerLink({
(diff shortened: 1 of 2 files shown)


paglias:
Have you tried putting the toggle button inside a `d-flex align-items-center` div like other icons on Habitica?

For example the streak icon is wrapped this way inside the `icons-right` div

```
          .icons-right.d-flex.justify-content-end
            .d-flex.align-items-center
              .svg-icon.streak(v-html="icons.streak")
```

8Mobius8:
@paglias Yep its in the same `d-flex align-items-center` div as the other icons.

toggl-alignment

I think the image needs to be resized but I wouldn't think that would be an simple option.

paglias:
@8Mobius8 sorry I missed that, to scale the image you coul try to use `transform: scale(ratio)` to the entire div: for example `transform: scale(0.7)` would make everything 70% of the initial size

8Mobius8:
@paglias Good idea! I tried instead of doing the whole `div` I decided to settle for just the toggl button and @ 85% and it looks much better now:

screen shot 2018-01-04 at 21 10 53
screen shot 2018-01-04 at 21 10 42

Alys:
I think that looks good!

dooart:
works fine! 👍

thank you :)

Alys:
@8Mobius8 Thanks so much for doing this! If you have a Habitica account, this PR can count towards your [Comrade](http://habitica.wikia.com/wiki/Contributor_Titles) [contributor tier](http://habitica.wikia.com/wiki/Contributor_Rewards). Please post your User ID here - you can find it by clicking on your avatar to view your profile; it's a string that looks like d904bd62-da08-416b-a816-ba797c9ee265.

PR #974, opened 2018-02-22
Fix toggl button inside the new Jira modal

### LINKED ISSUE #905: Support new Jira redesign
### Environment details

* Operation system: Mac OS
* Extension version: 1.7.0

### Expected behavior
Show Toggl button

### Actual behavior
Toggl button isn't displayed anymore

Open issue

### Additional details
Due to the removed sidebar in the redsign of Jira the Toggl button isn't displayed anymore. This needs to be shown in the newly added popup when you click on an issue.

jira

See [https://confluence.atlassian.com/jirasoftwarecloud/introducing-your-new-jira-experience-937886012.html](https://confluence.atlassian.com/jirasoftwarecloud/introducing-your-new-jira-experience-937886012.html)

dpsxp:
So, this should handle the new Jira design.

image

this should close #905

### DIFF
--- a/src/scripts/content/atlassian.js
+++ b/src/scripts/content/atlassian.js
@@ -28,14 +28,18 @@ togglbutton.render('#ghx-detail-view [spacing] h1:not(.toggl)', {observe: true},
   numElem.parentNode.appendChild(container);
 });
 
-// Jira 2017 sprint modal
-togglbutton.render('#ghx-issue-fragment [spacing] a img:not(.toggl)', {observe: true}, function () {
+
+// Jira 2018 sprint modal
+// Using the h1 as selector to make sure that it will only try to render the button
+// after Jira has fully rendered the modal content
+togglbutton.render('div[role="dialog"] h1:not(.toggl)', {observe: true}, function (titleElem) {
+  // need to all the way up until we find the modal wrapper
+  var rootElem = titleElem.offsetParent.offsetParent.offsetParent.offsetParent
+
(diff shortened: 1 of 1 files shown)


src/scripts/content/atlassian.js
```diff
@@ -28,14 +28,18 @@ togglbutton.render('#ghx-detail-view [spacing] h1:not(.toggl)', {observe: true},
...
 
-// Jira 2017 sprint modal
-togglbutton.render('#ghx-issue-fragment [spacing] a img:not(.toggl)', {observe: true}, function () {
+
+// Jira 2018 sprint modal
+// Using the h1 as selector to make sure that it will only try to render the button
+// after Jira has fully rendered the modal content
+togglbutton.render('div[role="dialog"] h1:not(.toggl)', {observe: true}, function (titleElem) {
+  // need to all the way up until we find the modal wrapper
+  var rootElem = titleElem.offsetParent.offsetParent.offsetParent.offsetParent
```
Phenome: Do we really need to use `offsetParent` instead of `parentNode`? `offsetParent` seems too dependant on elements being positioned.

dpsxp: The idea is to get to the dialog root element as fast as we can. I could change to use CSS selectors instead but looks like Jira is always changing the elements classes. :cry:

dooart:
works 👍 

hopefully this solution will keep this working for a longer while now

Issue #937, opened 2017-12-05
Persistent Pomodoro notifications


tekvsakdan:
It would be great if we could set that the Pomodoro notifications are persistent until we click on them. Persistent notifications are available since Chrome 50.

The reason is that we often missed the sound and the notification.

PR #976, opened 2018-03-01
The pomodoro notification should remain visible on screen

tekvsakdan:
The pomodoro notification should remain visible on screen until user activates or dismisses the notification.
It fixes issue #937.


shantanuraj:
Fixed in 794b2ef7e8b567d9c0deb9485875c055c66a3a77

Issue #966, opened 2018-02-01
Option to disable keyboard shortcuts


fonorobert:
From the user in case/637747:

> when working in Google Chrome on macOS, I use the keyboard shortcut cmd + shift + e for a function of a web app in Chrome.
> However, the toggl add-on is "stealing" this shortcut and start/stopping the current timer! How can I avoid that?
> Can I either disable this shortcut for toggl or disable all keyboard shortcuts for toggl in Chrome?

LulzAugusto:
@fonorobert chrome users can manage extensions' shortcuts (including Toggl) at chrome://extensions/shortcuts (this a URL).

PR #996, opened 2018-03-25
Adds Keyboard Shortcut Disabling

clptrsn:
Adds an option to the settings page to disable the keyboard commands.

This enhancement was requested here #966

image

### DIFF
--- a/src/html/settings.html
+++ b/src/html/settings.html
@@ -125,6 +125,11 @@
                 </p>
               </div>
             </li>
+            <li class="disable-keyboard-shortcut">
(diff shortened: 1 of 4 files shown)


LulzAugusto:
Hi, @clptrsn! 👋Thanks for taking the time implementing this, but since Chrome has its own way to manage extensions' shortcuts I don't think we need this option as of now, specially because it would confuse Firefox users.

LulzAugusto:
Oh, and in case you're interested, the way of doing it in chrome is by going to chrome://extensions/shortcuts. Cheers!

PR #556, opened 2016-05-21
Add feature for custom formatting the description text

sebastianwahn:
With this feature, it's possible to add a custom formatting to the description text which is submitted to the toggl website.

The current description format looks like:
"TaskId Description" if a task-id and a description is provided (JIRA).

With this feature, a user can set his own formatting. 
The formatting can be activated in the options. Afterwards a user can set the formatting with the currently provided tokens ($$taskId$$ and $$description$$).

A customized format can look like this now:
"$$taskId$ - $$description$$ - test"

erweiterungen

This will be replaced, so the final description will look like (assuming the taskID is "ABC-123" and description is "Some fancy description"):
"ABC-123 - Some fancy description - test"

timer_ _toggl

Currently only the jira-plugin is adjusted to support this feature, since some other sites do not have task-id's  (trello).
To support the formatting for other sites, the member "taskId" must be set on the object passed to createTimerLink()
banners_and_alerts_und_common_js_-_toggl-button_-____documents_workspace_toggl-button_

PR #998, opened 2018-03-26
Add work item ID in Toggl description

mrksoftware:
It is very useful to have the id of the work item in the toggl description.
In this way, when we start the timer, the description will have this format: "[xxxyyy] - workitem title"

### DIFF
--- a/src/scripts/content/visualstudio.js
+++ b/src/scripts/content/visualstudio.js
@@ -5,7 +5,7 @@
 // Individual Work Item & Backlog page
 togglbutton.render('.witform-layout-content-container:not(.toggl)', {observe: true}, function () {
   var link,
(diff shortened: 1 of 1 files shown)


dooart:
Hey, @mrksoftware. Thank you for the PR!

It works fine, but I'd like to ask you to change something: our convention for ids inside descriptions doesn't include a separator character. So it'd be `id title` for your example, eg: `110 fixing bug`.

We'll eventually continue work on https://github.com/toggl/toggl-button/pull/556 that should allow you to configure a custom description pattern.

mrksoftware:
Hi @dooart, you're welcome!  

I've changed the template as requested. Looking forward to #556 since our standard is `[id] - title` 😄

dooart:
Thank you! We'll release a new version soon with your fix.

PR #999, opened 2018-03-27
Fix Sync Issue with Entry being Deleted

### LINKED ISSUE #993: Discarding an entry on another client doesn't sync to Button
### Environment details
Both Chrome and Firefox on OSX

### Expected behavior
When I discard an entry on the web app, Toggl Button should sync and refresh the UI back to the idle state.

### Actual behavior
I need to click on the taskbar button to force a sync.

1. Start an entry with Toggl Button
2. Discard the entry on the web or desktop app
3. See that the Toggl Button UI still thinks the entry is running

clptrsn:
The extention should now update the icon when the currently running
timer is deleted. It will also set that deleted entry that was running
in the latest stopped entry for the resume shortcut.

This should fix #993

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -315,6 +315,11 @@ var TogglButton = {
         entry = null;
       }
       TogglButton.updateTriggers(entry);
+    } else if (data.action === "delete") {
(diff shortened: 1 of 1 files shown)


dooart:
Thank you for your contribution, @clptrsn! Working fine in Chrome and Firefox. 👍

PR #1001, opened 2018-03-27
Microsoft To-Do support

tsvetelin-pavlov:
Added support for Toggl button on Microsoft To-Do site:
screenshot-18-03-28-00-23-25
Since it is developed by the team of Wunderlist, it is very similar. The main difference is on the screen of the new functionality "My Day", that was missing in Wunderlist.

### DIFF
--- a/src/scripts/content/microsoft-to-do.js
+++ b/src/scripts/content/microsoft-to-do.js
@@ -1,0 +1,29 @@
+/*jslint indent: 2 */
+/*global $: false, document: false, togglbutton: false, createTag:false*/
+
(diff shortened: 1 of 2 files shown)


shantanuraj:
Looks good to me

My only gripe is that it would be better if the indent is set to 2 spaces.
I've created a [patch file](https://pastebin.com/raw/Rg5VSTQT) to get this merged quick.

You can apply the patch update the PR and it should be good to be merged

shantanuraj:
Closed in favour of #1033 
Thanks!

Issue #1010, opened 2018-04-11
Targetprocess update


kyllikekimmel:
case/675318
> After recent Targetprocess update and slight changes in its UI your Toggl Chrome Extension stopped appearing in Entity Details view in our application.

vedr4n:
I've got another user mentioning the same problem. 

Case 676088.

bearoutthere:
I can confirm this problem.  Toggl button is no longer visible on User Stories and Bugs.  The only workaround is to use right click menu to start time tracking.

S3ns3o:
Hello, I'm also using the button with target process, I'm on a fix, hope to push it quickly.

PR #1017, opened 2018-04-17
Fix targetProcess new gui version + add some crash protection

S3ns3o:
Hello,

A little fix consisting in trying both version of TargetProcess Gui to generate the button.

This could respond to  #1010.


shantanuraj:
Superseded by #1035 
Thanks @S3ns3o

PR #1025, opened 2018-04-24
Fix displaying timer on Jira board cards

### LINKED ISSUE #992: Button Showing Several Times On Same Page

### Environment details

* Operation system: Windows
* Extension version: 1.10.3

### Actual behavior: 
image

### Steps to reproduce the behaviour: 
- Browse around in Jida, reopen a task and there will be 3 buttons present, browse some more and re-open the same task and there will be 7 time tracking buttons for the same issue.

CASE #661226

Let me know if you need anything else.

langri-sha:
Accounts for additional H1 elements that might be present in the card description. Previously, adding `# Some title` in the description would break the card.

Fixes #992.


shantanuraj:
Screenshots for the view state on

#### Page reload with card open

image

#### Open card on current page

image

langri-sha:
Not really proud about the solution, but I picked something that's initially a placeholder and resolves to actual content to sync up.

PR #1028, opened 2018-04-26
Fix the older new Asana UI

dooart:
Apparently Asana is slowly rolling out UI changes and the recent fix actually broke the UI for people without the newest stuff.


dooart:
tested by filip 🚀

langri-sha:
LGTM :+1: :ship: :it:

PR #1038, opened 2018-05-21
Add install beacon to notify extension installed

rylek90:
Add a message listener to respond with extension version as a safe-guard, to the previous `window.TogglButton` beacon.

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -1714,3 +1714,12 @@ chrome.commands.onCommand.addListener(function (command) {
     }
   }
 });
+
+if (!FF) {
+  chrome.runtime.onMessageExternal.addListener(function handleVersionMessage (request, sender, sendResponse) {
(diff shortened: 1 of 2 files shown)


src/scripts/background.js
```diff
@@ -1714,3 +1714,16 @@ chrome.commands.onCommand.addListener(function (command) {
     }
   }
 });
+
+function handleVersionMessage (request, sender, sendResponse) {
+  if (request && request.message && request.message === 'version') {
+    sendResponse({ version: chrome.runtime.getManifest().version });
```
shantanuraj: Even though it should work, we might want to use `browser.runtime` here for Firefox.
Maybe store it 

```
const runtime = FF ? browser.runtime : chrome.runtime
```

and then is use so we don't have to do the check multiple times

rylek90: done

PR #1050, opened 2018-06-03
Added workast integration

rudivs:
- Toggl button appears in task lists
- Toggl button appears in subtasks
- Toggl button does not appear on lists of completed tasks
- Project is taken from the title of the list in which the task appears (this can be inconsistent between different organizing pages)

toggl_workast

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ Add Toggl one-click time tracking to popular web tools.
 - [VersionOne](https://versionone.com/)
 - [Waffle](https://waffle.io/)
 - [WordPress](https://wordpress.com)
+- [Workast](https://workast.io)
 - [Workfront](https://www.workfront.com/)
 - [Worksection](http://worksection.com/)
 - [Wrike](https://www.wrike.com/)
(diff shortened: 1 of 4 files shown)


rudivs:
That's a great suggestion, @Phenome, that was bothering me too but I couldn't think of a good solution. I also added some extra functionality; this is a summary of the changes:

- Get project from dashboard if the name of the current list is 'Open tasks'
- If the project name is a Slack channel (starts with `#`), strip the `#`, replace `_` and `-` with spaces, and make it TitleCase
- Get tags from Workast, whether the task is minimized or open in task details view

rudivs:
I just picked up a bug with the tags - any assignees are also picked up by the item-name selector. I'll see how I can make it more robust.

rudivs:
Tested and should be working now. Also cleaned up how the sub-project (list header) is accessed, and explicitly checks that the task is not completed before adding the Toggl button. This also clears some JavaScript errors that I noticed from completed tasks having the subProject being null.

Phenome:
Thanks, @rudivs ! It'll be available next release.

rudivs:
Thanks for approving!

PR #1051, opened 2018-06-05
Update git clone command in Readme

Lubits:
The old git clone url wasn't working for me. This fixes that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ https://addons.mozilla.org/en-US/firefox/addon/toggl-button-time-tracker/
 
 ## Installing from Source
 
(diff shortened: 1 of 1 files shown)


langri-sha:
Thanks for pointing it out! 

Can we change it to `git clone git@github.com:toggl/toggl-button.git`? My reasoning is that most of us do not have our credentials set up for authenticating via HTTP when we want to contribute back PRs :thinking:.

Lubits:
@langri-sha I agree, I have changed it to the SSH command

langri-sha:
Thank you very much @Lubits :raised_hands:!

PR #1056, opened 2018-06-08
Teamwork Projects UI FIX

Neayto:
The extension wasn´t working anymore because of HTML changes we made in Teamwork Projects, this PR is changing the currently used selectors.

### DIFF
--- a/src/scripts/content/teamwork.js
+++ b/src/scripts/content/teamwork.js
@@ -9,7 +9,7 @@ togglbutton.render('div.taskRHS:not(.toggl), div.row-rightElements:not(.toggl)',
     isTKO = false,
     className = 'huh',
     container = $('.taskIcons', elem),
-    project = $(".w-header-titles__project-name a").textContent;
+    project = $(".w-header-titles__project-name a").textContent.trim();
 
   if (container === null) {
     // check if TKO container is there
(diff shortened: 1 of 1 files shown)


src/scripts/content/teamwork.js
```diff
@@ -9,7 +9,7 @@ togglbutton.render('div.taskRHS:not(.toggl), div.row-rightElements:not(.toggl)',
     isTKO = false,
     className = 'huh',
     container = $('.taskIcons', elem),
-    project = $("#projectName > span").textContent;
+    project = $(".w-header-titles__project-name a").textContent;
```
Phenome: Hey Neyato,
Thanks for contributing.
This apparently needs a `trim()` otherwise it won't match the project names on Toggl, because of the extra spaces.

src/scripts/content/teamwork.js
```diff
@@ -86,7 +86,7 @@ togglbutton.render('div.taskRHS:not(.toggl), div.row-rightElements:not(.toggl)',
 // Tasks Detail View Page
 togglbutton.render('div#Task div.titleHolder ul.options:not(.toggl), .view-header ul.task-details-options:not(.toggl)', {observe: true}, function (elem) {
```
Phenome: Can you confirm if this whole section is still needed? I think I got to the "Task Detail View Page" but this selector doesn't trigger. Only the selector on "Tasks listing page in project" triggers again, so perhaps both pages can use the same selector/code? It seems to be working.

src/scripts/content/teamwork.js
Neayto: Ok thanks for the information. I changed it now.

Phenome:
This now lgtm 👍 . It'll be available next release!

PR #1063, opened 2018-06-21
Updates Slack integration

### LINKED ISSUE #601: Slack integration fix 2
See comments in  #555  that was closed.

### LINKED ISSUE #1002: Toggl button and Slack
Minor bug - When starting a Toggl button timer on a public slack channel, everything works as expected, however when the timer is started for a private channel or direct message channel, the first letter of the channel/direct message is cut off in the Toggl button description.

Tested and replicated with Chrome

rudivs:
This update fixes the issue with the first letter being removed from a private channel. However, it also changes the current behaviour of the Toggl button in Slack. I think it's an improvement, but that may just be my opinion. Changes included here are:

- Project is now taken from the channel name ***not team name as before***
- Hyphens (`-`) and underscores (`_`) are converted to spaces
- Project name is converted to Title Case
- Task description is left undefined for the channel title button
- Adds a Toggl button to the hover message actions list
- Task description for this button is taken from the message body
- Fixes #1002 
- Fixes #601

image

### DIFF
--- a/src/scripts/content/slack.js
+++ b/src/scripts/content/slack.js
@@ -1,12 +1,12 @@
 /*jslint indent: 2 */
 /*global $: false, document: false, togglbutton: false, createTag:false*/
-'use strict';
 
 togglbutton.render('#channel_name:not(.toggl)', {observe: true}, function () {
+  'use strict';
   var link,
     placeholder = $('.channel_title_info'),
-    description = $("#channel_name").textContent.trim().substr(1),
-    project = $('#team_name').textContent;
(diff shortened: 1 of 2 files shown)


rudivs:
Thanks @nunofmn I reverted to the default behaviour (although the description in the hover message is taken from the message body, which I think is logical from the context). Also squashed and rebased.

nunofmn:
Yes the behaviour in the hover message is fine 👍

LGTM 👍 

Thank you for you contribution. 🎉

Issue #1044, opened 2018-05-30
Problems with the TickTick integration


vedr4n:
### Environment details

* Operation system: High Sierra 10.13.4
* Extension version: Latest

### Expected behavior
Toggl Button should copy the task name from TickTick into the Description field. 

### Actual behavior
Toggl Button copies the task name + the words "What needs doing?". This is the default description/placeholder in the TickTick app. Toggl Button ignored this until 1 week ago.

### Additional details
TickTick made some changes to the app recently. It's possible those have caused the bug
https://ticktickblog.tumblr.com/post/174007155444/newtagging

**Note**: I haven't reproduced this myself, don't have a TickTick account. 

---
Case: 708218

PowerSnail:
I have a TickTick account, and I can reproduce this bug. 

If I may speculate the reason, I think it might relate to the following html structure:

```html
<div class="tasktitle" ...>
  <div class="... task-title ...">Actual Title Text</div>
  <div class="input-placeholder ...">...</div>
</div>
```

If Toggl is taking the text from `tasktitle`, it will include the extra "What needs doing?". Instead, if Toggl takes the text from `task-title`, it will include only the actual title text.

If I am wrong about how Toggl gets the text, please just ignore this.

Thanks.

PowerSnail:
Hi, I created a pull request to fix this #1064.

PowerSnail:
@vedr4n Hi, any chance of merging this change soon? 

Just a one line replacement.

vedr4n:
@PowerSnail Yup, everything is pretty much ready. Should happen early next week.

PowerSnail:
Thanks!

langri-sha:
Fixed via #1064.

PR #1064, opened 2018-06-26
fix the inclusion of placeholder text on ticktick.com

PowerSnail:
This is a fix for issue #1044.


langri-sha:
Works perfectly now. Thanks :bowing_man:!

PR #1069, opened 2018-07-02
Fix asana board task view

### LINKED ISSUE #1068: Asana integration is broken for board task view
From support

> The Toggl button is appearing in the issue detail in a list view and it does appear in the card summary in the board view but if I open a specific card the toggl button does not appear in the card detail.

See Board task view https://github.com/toggl/toggl-button/pull/1021

image

The button should show up in the "Mark Complete" row right as the last icon

The extension reports container to be `null` while attempting to the add the button to DOM

shantanuraj:
Closes #1068

## Testing

* The button should appear in board-card detail view

image

* The button should appear in the standard task detail view

image

(Right before the close/'X' button)

## Changes

* Fixed integration
* Made the button appear before the close button

### DIFF
--- a/src/scripts/content/asana.js
+++ b/src/scripts/content/asana.js
@@ -128,7 +128,7 @@ togglbutton.render('.SingleTaskPane-titleRow:not(.toggl)', {observe: true}, func
     return;
   }
   var link,
-    container = $('.SingleTaskPaneToolbarEasyCompletion', elem.parentNode),
+    container = $('.SingleTaskPaneToolbar'),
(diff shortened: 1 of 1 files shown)


dpsxp:
The icon is there, but why the color is different? The new one is always gray when not active. 🤔

shantanuraj:
@dpsxp see here https://github.com/toggl/toggl-button/blob/fix-asana-integration/src/scripts/content/asana.js#L139

we set the button to be gray to fit more with the toolbar icons aesthetic.

Issue #1053, opened 2018-06-06
Toggl button Trello/Jira issues


nicktoggl:
Some reports that starting Toggl button in Trello is causing crashes/lockups/isn't working

"I am trying to start a timer from a card in trello, i can input the project
(so it does link to toggle) but when i click the 'Done' button nothing
happens.
Even if i close the card the 'start timer' popup window hangs there
uncloseable and i can't do anything but refresh my browser.

Chrome 1.10.6"
/case/717583

"When using the Trello extension, the extension crashes and does not begin timing. Chrome latest version."

case/717865

VeljkoTN:
Could this be the reason:

> For me, it works. My Chrome language is set to English, however, my teams
Chrome is set to Chinese and they cannot get it to work with Trello or
Google Docs.

shantanuraj:
Set my language in Chrome and Trello to Chinese, added some title to the board

![image](https://user-images.githubusercontent.com/1716853/41039190-50ab9e38-69b6-11e8-9a68-40c4202cc82f.png)

Works fine for me https://streamable.com/94moz

Michelle112:
Nick requested that the user provide more info. These are the users extensions;
![gonzalo extensions](https://user-images.githubusercontent.com/19910107/41304452-d0dda762-6e35-11e8-917e-d2272121c536.png)

He also says he's using Chome Version 67.0.3396.79 (Build Official) (64 bits) 
Windows 10 pro 64 bits

shantanuraj:
Installed all the extensions from the screenshot, enabled them on my test card; the button still [works](https://streamable.com/3yqai)

Can we request the user to try the button with the other extensions disabled? If the button still doesn't work, send us a screen-recording with the browser console open.

ricotheque:
@shantanuraj The user still encountered problems unfortunately.

They were able to get a screenshot with a console error. Please check the case on Desk, and let us know if you need access to that. :)

maritoggl:
JIRA user having a similar issue - case/725339, also includes a screenshot of the error in console.

maritoggl:
Another one - case/729276

PR #1070, opened 2018-07-04
Fix Jira integration parentElement null error

shantanuraj:
Related issue: #1053

Desk case: case/725339

#1053 seems to be a common issue between trello and jira users and is probably related to the backend of the extension db/background scripts.

This PR doesn't address that issue specifically but is a fix for the most common error on [bugsnag](https://app.bugsnag.com/toggl-dot-com/toggl-button/errors/5b2ac4200ecf63001722ae2a?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[search][0][type]=eq&filters[search][0][value]=atlassian)

### DIFF
--- a/src/scripts/content/atlassian.js
+++ b/src/scripts/content/atlassian.js
@@ -81,7 +81,10 @@ togglbutton.render('.issue-header-content:not(.toggl)', {observe: true}, functio
     projectName: projectElem && projectElem.textContent.trim()
   });
 
-  $('.issue-link').parentElement.appendChild(link);
+  link.style.marginLeft = '8px';
+
+  var issueLinkContainer = $('.issue-link').parentElement || $('.aui-nav li').lastElementChild;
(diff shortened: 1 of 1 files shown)


Phenome:
Code looks OK and will probably fix it. But can you give me some steps to repro, so I can verify?
I couldn't, on `master`.

shantanuraj:
Hey!

So previously we were using this `'.issue-link')` selector.
It seems JIRA is doing some A/B testing with different markups so I added `'.aui-nav li'` as a fallback in case the former is not matched.

Also previously we were using

```javascript
$('.issue-link').parentElement.appendChild(link);
```

without null checking the result of our query, have fixed that in this commit.

For the repro part, sadly I am not getting the alternate JIRA layout myself but on bugsnag (linked in the description) you can find a lot of events with the errors stemming out from the previous code.

Issue #1072, opened 2018-07-04
www.meistertask.com not working with Toggl Button


TIIUNDER:
Hi guys,

I'm trying to use the Toggl button in Chrome with Meistertask, but it's not working. The button doesn't appear in Meistertask.

### Environment details

* Operation system: Ubuntu 16.04 / Google Chrome
* Extension version: 1.10.7

### Expected behavior

Toggl button appears in Meistertask

### Actual behavior

Toggl button doesn't appear in Meistertask

### Steps to reproduce the behavior

- Install Toggl button in Google Chrome
- Setup Toggl button to allow all or at least www.meistertask.com
- Open Meistertask task

### Additional details

I get the following error in the Google Chrome console:

`meistertask.js:10 Uncaught TypeError: Cannot read property 'textContent' of null
    at meistertask.js:10
    at Object.renderTo (common.js:146)
    at MutationObserver.<anonymous> (common.js:128)`

The code doesn't work here:

![screenshot_2018-07-04-16 53 03-1](https://user-images.githubusercontent.com/1225378/42283944-4879d288-7fab-11e8-8527-c6ff87ffc733.png)

TIIUNDER:
I tested a little bit and in this file is the issue:  `https://github.com/toggl/toggl-button/blob/master/src/scripts/content/meistertask.js`

In line 10 the project variable can not be set, because it doesn't find the title element. I removed this line and the button working, but without project.

TIIUNDER:
Fixed the project name and created the following pull request: #1075

PR #1075, opened 2018-07-06
Update meistertask.js

TIIUNDER:
Fix meistertask project name


TIIUNDER:
This should fix issue #1072

PR #1078, opened 2018-07-12
[add] backlogtool.com

mitakeck:


bryanph:
@mitakeck could you elaborate on why we need to add this? It redirects to `backlog.com`.

bryanph:
@mitakeck we recently fixed backlog integration again, see https://github.com/toggl/toggl-button/pull/1194. It might take a few days to release.

PR #1090, opened 2018-07-29
Package updates

langri-sha:
Adds rudimentary support for webpack, with some initial JS modules extracted.


langri-sha:
`API_URL` updated to address https://github.com/toggl/toggl-button/issues/1108.

PR #1160, opened 2018-10-21
Add support for new YouTrack layout

adinadin111:
YouTrack has recently introduced new layout for single issues. Since YouTrack 2018.3 the ability to choose between "old" and "new" layout for single issues was removed, now only the "new" layout is available.

This pull request adds support for this new layout. 

The toggl button is located on the right:
image
image


bryanph:
build this

bryanph:
sorry I had the wrong page up, checked the changes looks good!

PR #1172, opened 2018-10-31
Hide permissions tab in Firefox

### LINKED ISSUE #1060: Firefox Add-On: Add Permissions Disable All Settings
## Updated Issue

Firefox Add-On currently lacks the following settings menu that's present in the Chrome Extension.

## Original Issue
Hi Toggl team,

Big fan of this extension, thanks for your work on it. 

This is a feature request; if **Issues** isn't an appropriate place for this, feel free to close.

I love having the button available in my browser, and really wish that the button was _not_ [integrated](https://github.com/toggl/toggl-button/wiki/Where-can-I-find-the-Button%3F) into the DOM.

I'd imagine that there are others who'd desire a setting that allows a user to turn off these integrations.

tcrammond:
We have a wildcard permission in Firefox, so the permissions tab is redundant.

It's possible to implement for Firefox, but requires some additional work around origin permissions. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/request

So we are hiding the permissions tab for now to reduce confusion, because it simply doesn't do anything in Firefox yet. The plan is to revisit this after resolving some more critical issues first. Part of #1060.

* Do not show Permissions tab link and tab content in Firefox
* Do not show Permissions instructional popup on Firefox
* Add a check when loading the active tab from local storage in Firefox, as there is one fewer tab now and there is a possibility of an error.
* Remove/clean up some conditional Firefox code in the settings script to ease any re-implementation later, as it can in fact use the same APIs as chrome. There's still conditions and such there, but I removed alternate branches that will not be needed later.

Notes for testing:
- Chrome functions the same as before.
- Firefox does not ever allow the Permissions tab to be shown, and otherwise functions without issue.

### DIFF
--- a/src/scripts/settings.js
+++ b/src/scripts/settings.js
@@ -240,8 +240,12 @@ var Settings = {
       disabled,
       checked,
       customs,
-      tmpkey,
-      keyc;
+      tmpkey;
+
+    if (FF) {
+      // Permissions settings are not available on Firefox at this time.
+      return;
+    }
 
     try {
       // Load Custom Permissions list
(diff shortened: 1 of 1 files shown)


bryanph:
can you remove the function() to function () changes? Makes it a bit easier to review

src/scripts/settings.js
```diff
@@ -642,9 +606,19 @@ document.addEventListener('DOMContentLoaded', function(e) {
     );
     Settings.$sendErrorReports = document.querySelector('#send-error-reports');
 
+    if (FF) {
```
bryanph: maybe add a comment here to indicate it has to do with hiding the permissions tab.

bryanph:
did you check out the `optional_permissions` in firefox? any success?

tcrammond:
@bryanph Yes, it does look like we'll be able to use a wildcard there after all to bring back everything including custom URLs (additional work in the codebase will still be needed). Other critical bug fixes are priority for the next release, though, so we'll revisit this after that.

Issue #1130, opened 2018-09-18
Chrome bug - lost time


barnabaskecskes:
### Environment details

* Operation system: MacOS High Sierra v10.13.6
* Chrome version: 69.0.3497.100
* Extension version: 1.11.1

### Expected behavior
If you have a timer going on and you overwrite the time, you expect the new value to be saved in the web app.

### Actual behavior
The real time that has passed is saved instead.

### Steps to reproduce the behavior
Start a timer in the extension. Modify the time and stop the clock. Go to your web app and see only a few seconds stored.

### Additional details
I have tried to _wait_ just in case slow it is due to slow network connection, but even when the web app in the background has already synchronised the clock, when I stop, the web app shows the right times, but the delta between shows only the seconds. When clicking, the times update automatically to reflect (the incorrect) elapsed time. (See below screenshots demonstrating the behaviour.)

### Screenshots
Waiting to synchronise with web app after modifying the timer:
<img width="410" alt="screen shot 2018-09-18 at 15 04 37" src="https://user-images.githubusercontent.com/3535449/45693139-5905cf00-bb54-11e8-9c95-3d49004068cd.png">

After synchronisation, stopping the timer:
<img width="464" alt="screen shot 2018-09-18 at 15 04 48" src="https://user-images.githubusercontent.com/3535449/45693145-5d31ec80-bb54-11e8-8599-383da2986d27.png">

The web app shows the right times, but incorrect delta:
<img width="1203" alt="screen shot 2018-09-18 at 15 05 00" src="https://user-images.githubusercontent.com/3535449/45693160-61f6a080-bb54-11e8-8788-2cb09c04366c.png">

Clicking on the times auto-adjust them to reflect the wrong delta:
<img width="342" alt="screen shot 2018-09-18 at 15 05 17" src="https://user-images.githubusercontent.com/3535449/45693164-6458fa80-bb54-11e8-8436-926c88c60f6b.png">

PR #1174, opened 2018-11-02
Fix wrong durations for entries that were manually edited

### LINKED ISSUE #1171: Bug when changing duration and stopping afterwards in Toggl Button
The bug was reported by one of our users and I was able to reproduce it following the steps below.

Steps to reproduce:
1. Start a new time entry or continue recent in Toggl Button
2. Click on the active entry
3. Change the time (replace 00 minutes with 10) and hit done (eg if you forgot to start tracking and want to add 10 minutes to the tracking)
4. The counter in Toggl Button correctly shows as if you have been tracking for 10+minutes
5. Even www.toggl.com shows running for 10minutes+
6. Open Toggl Button and click 'stop'

Problem:
The entry gets saved but not with 10minutes+ but with the actual duration that it took to complete the previous steps, eg less than 1 minute

What's expected:
The entry gets stored as it was displayed with 10minutes+

Note: This only happens if the entry is stopped in the Toggl Button. If everything is done the same but the entry is stopped on toggl.com then it works as expected

dan-iway:
Toggl-Button does not update the duration when updating the start date. This results in entries that have a different duration than what was displayed in toggl-button and in toggl.com as well.

This commit updates duration as well whenever the start time changes to fix this.

Fixes #1171 and #1130

### DIFF
--- a/src/scripts/popup.js
+++ b/src/scripts/popup.js
@@ -451,8 +451,7 @@ window.PopUp = {
   getStart: function() {
     var arr = document.querySelector('#toggl-button-duration').value.split(':'),
       duration,
-      now,
-      start;
+      now;
 
     if (!PopUp.isNumber(arr.join(''))) {
       return false;
@@ -461,8 +460,7 @@ window.PopUp = {
     duration = 1000 * (arr[2] || 0) + 60000 * arr[1] + 3600000 * arr[0];
 
     now = new Date();
-    start = new Date(now.getTime() - duration);
-    return start.toISOString();
+    return new Date(now.getTime() - duration);
   },
 
   isNumber: function(n) {
@@ -502,7 +500,8 @@ window.PopUp = {
       start = PopUp.getStart();
 
     if (start) {
-      request.start = start;
+      request.start = start.toISOString();
+      request.duration = -1 * Math.floor(start.getTime() / 1000);
     }
 
     PopUp.sendMessage(request);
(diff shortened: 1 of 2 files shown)


src/scripts/popup.js
```diff
@@ -502,7 +500,8 @@ window.PopUp = {
       start = PopUp.getStart();
 
     if (start) {
-      request.start = start;
+      request.start = start.toISOString();
+      request.duration = -1 * Math.floor(start.getTime() / 1000);
```
bryanph: what do you base this value on? :thinking:

dan-iway: https://github.com/toggl/toggl_api_docs/blob/master/chapters/time_entries.md

> duration: time entry duration in seconds. If the time entry is currently running, the duration attribute contains a negative value, denoting the start of the time entry in seconds since epoch (Jan 1 1970). The correct duration can be calculated as current_time + duration, where current_time is the current time in seconds since epoch. (integer, required)

dan-iway: For currently running time entries the duration needs to be set to negative start timestamp. This was not done before, so the duration never gets updated and toggl-button uses the old duration timestamp on the stop event which results in entries that can read 

_start: 10:42, stop 10:53, duration 11 seconds_

bryanph: great, thanks for educating me :smile:

PR #1177, opened 2018-11-05
allow setting custom domains and api url through env

bryanph:
This allows the toggl team to set `API_URL` and `TOGGL_API_HOST` as environment variables to allow for testing toggl-button locally using the super uber duper secret URL.


webpack.config.js
```diff
@@ -2,7 +2,7 @@ const path = require('path');
 const CleanPlugin = require('clean-webpack-plugin');
 const CopyPlugin = require('copy-webpack-plugin');
 const FileManagerPlugin = require('filemanager-webpack-plugin');
-const { DefinePlugin } = require('webpack');
+const { DefinePlugin, EnvironmentPlugin } = require('webpack');
```
Phenome: I think we can ditch `DefinePlugin` here.

bryanph: thanks, fixed that

webpack.config.js
```diff
@@ -81,12 +79,12 @@ module.exports = config(({ development, production, version }) => ({
...
         from: 'chrome-manifest.json',
         to: 'chrome/manifest.json',
-        transform: addVersion
+        transform: transformChromeManifest
       },
       {
         from: 'firefox-manifest.json',
         to: 'firefox/manifest.json',
-        transform: addVersion
+        transform: transformFirefoxManifest
```
shantanuraj: You can probably reuse a significant amount of code by taking an additional argument in this function

```
transformManifest(browser: 'firefox' | 'chrome')
```

Then in the function just conditionally add the chrome specific permissions by checking `browser` variable.

webpack.config.js
```diff
@@ -132,24 +130,36 @@ function copy(o) {
...
-    JSON.stringify(
-      {
-        version: pkg.version,
-        ...JSON.parse(content.toString())
-      },
-      undefined,
-      2
-    )
-  );
+function transformChromeManifest(content) {
```
shantanuraj: Rewriting this as you make DRY this code 😄 

```javascript
function transformManifest(browser) {
  return function (context) {
    // ...
    // ...
    if (browser === 'chrome') {
      manifest.externally_connectable.matches = [...]
    }
  }
}
```

bryanph: good point! I updated the PR thanks.

bryanph:
@shantanuraj thanks, i integrated your suggestions

Issue #1170, opened 2018-10-30
Closing the notification popups saying "click here to...." still execute the action


atleta:
I've found this happening in two situations (there may be more):
* when "remind me to track" is enabled
* when "idle detection" is enabled

The extension will show me a popup when/if the condition for the above is triggered (e.g. I'm idle while I'm tracking time OR I'm active while I'm not tracking time). The popup will say that I should click it to do some action (discard idle time or start tracking). Now I'd expect to be able to prevent this from happening if I just close the popup with the [X] button, but even if I do that, the action will be executed as if I clicked the popup.

### Environment details

* Operation system: Ubuntu 17.04
* Extension version: 1.12.0
* Browser: Firefox 57.0.4

### Expected behavior
Close popup without executing the action I'm being reminded of/being asked about by the popup. (E.g. discard idle time.)

### Actual behavior
It does it anyway. (I.e. it will discard the idle time even if I close the popup instead of clicking it.)

### Steps to reproduce the behavior
Enable "idle detection" and/or "remind me to track" and be idle or be active (depending on which one you want to test) then click the [X] icon on the popup.
### Additional details

* screenshots or screencapture

tcrammond:
Hi, thanks for reporting this issue. Hoping to have this one resolved in the next release.

related to #778, #1116.

tcrammond:
Fixed and released in `1.13.0`👌.

Issue #1116, opened 2018-08-30
Timer starts automatically even though setting for starting automatically is unchecked


nhanisch:
### Environment details

* Operation system: Linux Debian 9
* Extension version: Firefox Toggl browser extension 1.10.7

### Expected behavior
Timer will only start if I explicitly click to start the timer through the browser extension or on Toggl.com. This behavior should only change if I change my setting in the Toggl settings to "Start automatically". 

### Actual behavior
Timer will occasionally start while I am using my web browser sometime shortly after beginning work. It is unclear when it starts because i am usually very focused and only catch it 30+ minutes into working.  

### Steps to reproduce the behavior
From a computer running Debian 9 and the Firefox Toggl extension 1.10.7, it seems to happen randomly. It does not start immediately after opening Firefox. I rarely close my browser at all, but I suddenly notice it will be running and be 30+ minutes into running.

tcrammond:
Hi @nhanisch. I know it's been a while, but have you seen this issue much since you first reported it? Do you know if the timer being started is related to a particular integration, or is it a blank entry?

nhanisch:
@tcrammond It always starts the timer with a blank entry. At this point my Toggl browser extension has updated to v1.12 and the issue is still occurring.

Occasionally the extension will sign me out, and so I open the Toggl web app and it automatically signs me back in on the extension. In addition to this, I occasionally open the Toggl web app to fix time entries or look at reports.

PR #1180, opened 2018-11-08
Fix notification actions triggering without user input

### LINKED ISSUE #778: Clicking on the X button inside the Reminder notification starts the timer
This issue is specific to Toggl Button for Firefox on Windows. On MacOS, the notification doesn't have an X button. 

When you click on the X (close) icon inside the notification the notification is closed and the timer starts. In other words, the behavior is the same as clicking anywhere else on the notification. 
x icon

---
Desk case: 430711

tcrammond:
Fix multiple issues arising from notifications triggering their action when closed, rather than only when the user interacts with the notification. The main offender was the timer starting automatically whenever the notification is shown.

See commit for details. Fixes #778, #1116, #1170.

### Recommendations for testing

I mainly tested the "reminder to track time". I set this to 1 minute, and started+stopped the timer to trigger the reminder behind the scenes.

When the notification shows, if you either don't do anything or choose "close" (not on macOS+Firefox), the timer **should not start**. 

It's the same for all notifications, though. Needs testing on both Chrome and Firefox, and we should test Windows Firefox.

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -721,7 +721,7 @@ window.TogglButton = {
                 })
               );
             }
-            TogglButton.triggerNotification();
+            TogglButton.setNannyTimer();
             ga.reportEvent(timeEntry.type, timeEntry.service);
             if (cb) {
               cb();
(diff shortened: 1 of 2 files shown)


dooart:
code lgtm

PR #1191, opened 2018-11-19
Re-enable permissions in Firefox

### LINKED ISSUE #1060: Firefox Add-On: Add Permissions Disable All Settings

### LINKED ISSUE #1169: Adding custom service doesn't work in Firefox
### Environment details

* Operation system: Arch Linux
* Extension version: 1.12.0

### Expected behavior

Able to add custom service.

### Actual behavior

Custom url adds with AgenoCRM setting only.

- Try to add new custom domain url permission
- Fail

image
image

tcrammond:
Previously permissions were broken and/or disabled since the start due to slight differences in the extension API, which seem to be cleared up now.

unfortunately the diff is quite bad. Closes #1060, closes #1169  and probably others that I'll find.

### Changes

- Show the permissions tab in Firefox again
- Change firefox manifest to no longer have wildcard permission by default
- Change firefox manifest to have a wildcard permission in *optional_permissions*, in the same vein as Chrome. This allows us to request permission on any origin at runtime (i.e. turn on/off domains and add custom domains).
- Remove redundant conditional Firefox code in the settings module, consolidating Chrome/Firefox code as they function exactly the same now.

Other:
- Extract each permissions event handlers to its own named function, and remove/add the listener on each new render rather than adding more and more duplicate listeners over time. Fixes awful issues in Firefox where clicking a button could cause progressively more listeners to be created. The root cause is buried somewhere in the way the settings page is rendered, this is a clean fix until such a time as we re-write it.
- Refactor permission filtering in a similar fashion to be a little more manageable.
- Minor html change to settings page layout and input

### Testing

This needs a bunch of testing and playing around with in Firefox, but also Chrome.

I managed to test a custom Jira domain by installing Jira server solution locally.. I wouldn't worry about testing custom domains again like that, just that you can add and remove them correctly in the settings.

### DIFF
--- a/src/scripts/settings.js
+++ b/src/scripts/settings.js
@@ -21,7 +21,6 @@ if (FF) {
 document.querySelector('#version').textContent = `(${process.env.VERSION})`;
 
 var Settings = {
-  eventsSet: false,
   $startAutomatically: null,
   $stopAutomatically: null,
   $showRightClickButton: null,
@@ -32,7 +31,6 @@ var Settings = {
   lastFilter: null,
   $permissionFilter: document.querySelector('#permission-filter'),
   $permissionFilterClear: document.querySelector('#filter-clear'),
-  permissionItems: [],
   $permissionsList: document.querySelector('#permissions-list'),
   $newPermission: document.querySelector('#new-permission'),
   $originsSelect: document.querySelector('#origins'),
@@ -242,11 +240,6 @@ var Settings = {
       customs,
       tmpkey;
 
-    if (FF) {
-      // Permissions settings are not available on Firefox at this time.
-      return;
-    }
-
     try {
       // Load Custom Permissions list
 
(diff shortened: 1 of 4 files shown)


bryanph:
Yep that's what I did. So I tried again... and this time the functionality does work. I think it might have been something about the confirmation window not showing up on my system for some reason. The only thing is that it still looks weird styling-wise. Here is a picture with the error logs:

image

src/scripts/settings.js
```diff
@@ -664,12 +638,14 @@ document.addEventListener('DOMContentLoaded', function(e) {
...
       }
       Settings.lastFilter = val;
-      for (key in Settings.permissionItems) {
-        if (Settings.permissionItems.hasOwnProperty(key)) {
-          if (Settings.permissionItems[key].id.indexOf(val) !== -1) {
-            Settings.permissionItems[key].classList.add('filter');
-          } else if (!!Settings.permissionItems[key].classList) {
-            Settings.permissionItems[key].classList.remove('filter');
+
+      var permissionItems = document.querySelectorAll('#permissions-list li');
```
shantanuraj: We can modernize some of the code a bit

```javascript
const permissionItems = [...document.querySelectorAll('#permissions-list li')];
permissonItems.forEach((item) => {
    //
});
```

begone `hasOwnProperty`

bryanph: I must say I actually prefer the original, it's pretty unclear what the spread does.

shantanuraj: Actually the `forEach` function is actually present on `NodeList`, so can use it directly.

The `...` converts the `NodeList<Element>` to an array of `Element`.
Can also be written as
```javascript
const permissionItems = Array.from(document.querySelectorAll('#permissions-list li');
```

tcrammond:
@bryanph can you inspect the settings' iframe, using this dropdown in the firefox debugger for the extension:

screen shot 2018-11-22 at 11 13 32


and see what happened to the html? should just be a list..

screen shot 2018-11-22 at 11 13 22

bryanph:
@tcrammond I changed the styles a bit so it looks better in firefox. There are still some overflow issues though but that seems to happen because in firefox the settings page doesn't have a viewport height for some reason. I suppose it is injected with a fixed height or something :man_shrugging: 

In firefox:
image


In chrome:
image

tcrammond:
@bryanph nice, some issues though

We've lost styling of the custom permissions list items:

**Before:**

screen shot 2018-11-23 at 15 02 09

**After:**

(buttons wrap on normal permissions too now, not really a deal breaker, mainly want to keep the list item styles)

screen shot 2018-11-23 at 14 58 33

We also have an issue with tooltip on wide viewport settings (read: Firefox, at least on macOS) again I think you may have undone some width limits on the inner content that I had set up for this reason (see commit `3b98d33 
Improve styles on wider width viewports (Firefox)`). We don't wanna have to reinvent the help text at this point

screen shot 2018-11-23 at 14 58 14

bryanph:
@tcrammond thanks, should be good now :crossed_fingers:

PR #1222, opened 2018-12-12
Fix breaking error in Todoist when project is missing

tcrammond:
Seen in error reporting, `closestEditor` could be undefined. This PR is a defensive fix so the integration stops breaking for users with uncaught error.

Will open issue to fix integration fully.


tcrammond:
Closing for now as #1224 should resolve this. Will follow up after.

PR #1225, opened 2018-12-12
Add setting to control automatic tagging

### LINKED ISSUE #1217: Auto-Adding of Todoist Labels as Tags as an Option?
### Environment details

* Operation system: Windows 10
* Extension version: 1.15.0
* Browser: Google Chrome
* App using Toggl Button: Todoist

Are there any plans to add an option to enable/disable the new automatic creation of Toggl Tags from Todoist Labels ( #1197 )? Since you need to explicitly tell the Toggl button for each time entry to NOT bring over the Label/Tags, people are having dozens of new tags created in Toggl every day for our company and we don't really want them in Toggl since everyone uses them differently.

tcrammond:
This setting controls whether service integrations will automatically add any tags/labels detected to the time entry (and therefore creating them in Toggl, too). It defaults to **disabled**.

The main reasoning stems from feedback in #1217, where users have their own tags in the integration - unrelated to the way the team tracks time -  and now the Toggl workspace is being polluted with dozens of new tags since automatic tagging was added to that integration.

This affects behaviour of **very few** other existing integrations (`<= 4` at best), so any unrest when this setting is released should be minimal.

Closes #1217.

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -436,6 +436,7 @@ window.TogglButton = {
       defaultProject = db.getDefaultProject(),
       rememberProjectPer = db.get('rememberProjectPer'),
       entry;
+    const enableAutoTagging = db.get('enableAutoTagging');
     TogglButton.$curService = (timeEntry || {}).service;
     TogglButton.$curURL = (timeEntry || {}).url;
 
@@ -463,7 +464,7 @@ window.TogglButton = {
       pid: timeEntry.pid || timeEntry.projectId || null,
       tid: timeEntry.tid || null,
       wid: timeEntry.wid || TogglButton.$user.default_wid,
-      tags: timeEntry.tags || null,
+      tags: enableAutoTagging ? (timeEntry.tags || null) : null,
       billable: timeEntry.billable || false,
       created_with: timeEntry.createdWith || TogglButton.$fullVersion
     };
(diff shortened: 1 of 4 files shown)


src/scripts/settings.js
```diff
@@ -598,6 +603,7 @@ document.addEventListener('DOMContentLoaded', function(e) {
       '#send-usage-statistics'
     );
     Settings.$sendErrorReports = document.querySelector('#send-error-reports');
+    Settings.$enableAutoTagging = document.querySelector('#enable-auto-tagging')
```
leafi: Missing semicolon (compared to rest of file)

src/scripts/background.js
```diff
@@ -436,6 +436,7 @@ window.TogglButton = {
       defaultProject = db.getDefaultProject(),
       rememberProjectPer = db.get('rememberProjectPer'),
       entry;
+    const enableAutoTagging = db.get('enableAutoTagging')
```
leafi: No semicolon again, but it looks more intentional here :thinking:

src/scripts/background.js
```diff
@@ -463,7 +464,7 @@ window.TogglButton = {
       pid: timeEntry.pid || timeEntry.projectId || null,
       tid: timeEntry.tid || null,
       wid: timeEntry.wid || TogglButton.$user.default_wid,
-      tags: timeEntry.tags || null,
+      tags: enableAutoTagging ? timeEntry.tags || null : null,
```
nunofmn: Not sure if it is related with the linter 🤷‍♂️ 
```suggestion
      tags: enableAutoTagging ? (timeEntry.tags || null) : null,
```
or
```suggestion
      tags: (enableAutoTagging && timeEntry.tags) || null
```

PR #1256, opened 2019-01-11
Wrike: Extract project name and issue number

archnode:

This adds further information to the description (wrike issue number) as well as the project name (folder / project name) to the toggle button integration.

### DIFF
--- a/src/scripts/content/wrike.js
+++ b/src/scripts/content/wrike.js
@@ -1,18 +1,42 @@
 'use strict';
 
-togglbutton.render('.wspace-task-view:not(.toggl)', { observe: true }, function (
-  elem
-) {
-  const container = $('.wrike-panel-header-toolbar', elem);
+togglbutton.render(
+  '.wspace-task-view:not(.toggl)',
+  { observe: true },
+  function (elem) {
+    const container = $('.wrike-panel-header-toolbar', elem);
 
-  const titleElem = function () {
-    return $('title').textContent.replace(' - Wrike', '');
-  };
+    const descriptionText = function () {
+      const urlString = window.location.href;
+      const url = new URL(urlString);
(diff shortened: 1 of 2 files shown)


archnode:
Thank you for testing and your screenshot! I extract the issue number from the url that gets set when the popup is opened. Maybe a cross-browser issue - Are you testing in Chrome?

tcrammond:
Testing in Chrome, that's right. Parameter `'ot'` doesn't seem to exist in my issue URLs. I was using a new account and did not pick any kind of workspace template during setup.

archnode:
I'm sorry for the long delay: The problem with the parameter stems from the different task views in wrike (major oversight, new to wrike myself).

I updated the pr with:
- Moved Issue number to back of description (As this isn't as important / canonical as in other systems)
- Make Issue number optional (This enables tracking of non-issues if presented in a task view (Projects in subfolders for example)
- Fixes extraction problem with different task views.

src/scripts/content/wrike.js
```diff
@@ -1,18 +1,45 @@
...
+        var hashParts = hash.split("&");
+        var taskId = null;
+        for (var i = 0; i < hashParts.length; i++) {
+          var partSplit = hashParts[i].split('=');
+          if (partSplit[0] == 'ot' || partSplit[0] == 't') {
+            taskId = partSplit[1];
+          }
+        }
+        // We try to extract a title
+        // var titleElem = $('.title-field-ghost', elem);
```
tcrammond: Please remove this if it's not needed 😇

archnode:
Thank you for the feedback. I made the changes you suggested - the button in the pop-up looks equally misaligned, so I added the margin generally.

archnode:
I tried my best to update the branch!

PR #1260, opened 2019-01-17
Fix continued entries not copying tags

### LINKED ISSUE #1259: Continuing a task does not copy over tags
* **OS version**:  Win 10 x64
* **Browser version**: Chrome 71.0.3578.98 (Official Build) (64-bit)
* **Extension version**: 1.18.0

### 🐛 Describe the bug

Continuing an entry from the button dropdown does not add tags that were included in the previous entry. This issue is limited to the extension, it does not occur in the web interface at https://toggl.com/app/timer.

### Expected behaviour

When continuing a previously made entry that contains a tag, the new entry should copy those tags.

### Steps to reproduce

1. Click extension button in top-right of browser window
2. Start a new entry with one tag selected
3. Stop the entry after a few seconds
4. Again click the extension button in top-right of browser window
5. Hover over the most recent entry and click the green continue button
6. Stop the entry after a few seconds
7. The new entry will not contain any tags, while the entry that it was based off of does

tcrammond:

Ensures tags are included when continuing a time entry from the toolbar popup.

This was a regression introduced by the "enable auto tagging" flag. Resolves #1259.

`createTimeEntry` is used by all methods of creating or updating an entry; and it's not always clear where the request comes from. In the relevant cases, I have ensured that the request `type` is following through and check against that to determine if tags are included.

The only time tags should **not** be included is an event unrelated to continuing entries:
* starting a new timer from a tag-supported integration (E.g Todoist) when the `Enable autotagging` setting is disabled.

We are fixing:
* "Continue latest" button in the toolbar popup
* Clicking "Continue" on an entry in the toolbar popup list

 Resolves #1259.

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -430,13 +430,15 @@ window.TogglButton = {
   },
 
   createTimeEntry: function(timeEntry, sendResponse) {
-    var project,
-      start = new Date(),
-      error = '',
-      defaultProject = db.getDefaultProject(),
-      rememberProjectPer = db.get('rememberProjectPer'),
-      entry;
+    const type = timeEntry.type;
+    const start = new Date();
+    const defaultProject = db.getDefaultProject();
+    const rememberProjectPer = db.get('rememberProjectPer');
     const enableAutoTagging = db.get('enableAutoTagging');
+    let entry;
+    let error = '';
+    let project;
+
     TogglButton.$curService = (timeEntry || {}).service;
     TogglButton.$curURL = (timeEntry || {}).url;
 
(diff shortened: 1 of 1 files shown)


src/scripts/background.js
```diff
@@ -430,13 +430,15 @@ window.TogglButton = {
   },
 
   createTimeEntry: function(timeEntry, sendResponse) {
-    var project,
-      start = new Date(),
-      error = '',
-      defaultProject = db.getDefaultProject(),
-      rememberProjectPer = db.get('rememberProjectPer'),
-      entry;
+    const type = timeEntry.type
```
shantanuraj: Missing a ';'
Though I am not sure if we require them or disallow semicolons

PR #1277, opened 2019-02-05
Fix todoist list re-renders

### LINKED ISSUE #1275: Performance regression for Todoist integration
* **OS version**: Ubuntu 18.04.1 LTS
* **Browser version**: google-chrome Version 71.0.3578.98 (Official Build) (64-bit)
* **Extension version**: 1.17.1

### Bug description

When selecting a parent project in Todoist that has nested sub-projects and many tasks, Todoist becomes unresponsive.  Via chrome developer tools I tracked this down to being due to toggl-button.  Disabling the extension resolves the problem.

I believe that this is a recent regression related to changes in Todoist version 935.  https://get.todoist.help/hc/en-us/articles/115005442125-Todoist-for-Web-Changelog

>  🐛 You may notice that Todoist runs more smoothly across the app thanks to some performance optimizations.

I'm pretty familiar with Todoist's behavior (I wrote a fairly large extension for it, [todoist-shortcuts](https://github.com/mgsloan/todoist-shortcuts)).  I believe that what's changed here is that now it creates the dom nodes for the tasks sequentially, possibly in batch, instead of creating them all at once.  Since toggl-button uses a mutation observer, it's seeing many more mutations than it used to, and re-rendering all of the toggl buttons.  So I'm guessing this results in `O(n^2)` runtime when the tasks get rendered, because toggl-button seems to modify every single task on every single dom mutation.

### Expected behavior

I expected Todoist + toggl-button to not hang when viewing many tasks.

### Steps to reproduce

1. Install toggl-button

2. Visit a todoist project with many tasks, particularly with nested projects

3. Observe that the browser goes out to lunch, profiling reveals it's very busy.

shantanuraj:

- Render button only if task is hovered on
- Remove button on leaving a task

- Enable the todoist integration
- toggl-button should render on task items on hover
- Timer actions should work as usual
- The page shouldn't freeze

Closes #1275 

@mgsloan can you check if this improves the situation for your use-case

### DIFF
--- a/src/scripts/content/todoist.js
+++ b/src/scripts/content/todoist.js
@@ -1,12 +1,16 @@
 'use strict';
 /* global togglbutton, $ */
 
-togglbutton.render(
-  '.task_item .content:not(.toggl)',
-  { observe: true },
-  elem => {
-    const container = $('.text', elem);
+const EDITOR_ID = 'editor';
+const TASK_ELEMENT = '.content';
+const TOGGL_BUTTON = '.toggl-button';
 
+function insertTogglButton (event) {
+  const { target } = event;
+  const shouldInsert = target.matches(TASK_ELEMENT) && !target.querySelector(TOGGL_BUTTON);
+
+  if (shouldInsert) {
+    const container = target.querySelector('.text');
     const descriptionSelector = () => {
       const clone = container.cloneNode(true);
       let i = 0;
@@ -40,23 +44,20 @@ togglbutton.render(
     };
 
     const tagsSelector = () => {
-      const tags = elem.querySelectorAll('.labels_holder a:not(.label_sep)');
-
-      return [...tags].map(tag => {
-        return tag.textContent;
-      });
+      const tags = target.querySelectorAll('.labels_holder a:not(.label_sep)');
+      return [...tags].map(tag => tag.textContent);
     };
 
     const link = togglbutton.createTimerLink({
       className: 'todoist',
       description: descriptionSelector,
-      projectName: getProjectNames(elem),
+      projectName: getProjectNames(target),
       tags: tagsSelector
     });
 
     container.insertBefore(link, container.lastChild);
   }
-);
+}
 
 /*
 Projects may have a hierarchy in Todoist.
@@ -71,7 +72,7 @@ E.g.
 - Toggl Button will first check if MyFeatureProject exists, and if it doesn't, try to use the next parent etc.
 */
 
-function getProjectNameFromLabel(elem) {
+function getProjectNameFromLabel (elem) {
   let projectLabel = '';
   const projectLabelEle = $('.project_item__name', elem.parentNode.parentNode);
   if (projectLabelEle) {
@@ -80,7 +81,7 @@ function getProjectNameFromLabel(elem) {
   return projectLabel;
 }
 
-function getParentEle(sidebarCurrentEle) {
+function getParentEle (sidebarCurrentEle) {
   const levelPattern = /(?:^|\s)indent_(\d*?)(?:\s|$)/;
   const curLevel = sidebarCurrentEle.className.match(levelPattern)[1];
   const parentClass = 'indent_' + (curLevel - 1);
@@ -95,11 +96,11 @@ function getParentEle(sidebarCurrentEle) {
   return parentCandidate;
 }
 
-function isTopLevelProject(sidebarCurrentEle) {
+function isTopLevelProject (sidebarCurrentEle) {
   return sidebarCurrentEle.classList.contains('indent_1');
 }
 
-function getProjectNameHierarchy(sidebarCurrentEle) {
+function getProjectNameHierarchy (sidebarCurrentEle) {
   const projectName = $(
     '.name',
     sidebarCurrentEle
@@ -113,11 +114,11 @@ function getProjectNameHierarchy(sidebarCurrentEle) {
   return [projectName].concat(getProjectNameHierarchy(parentProjectEle));
 }
 
-function projectWasJustCreated(projectId) {
+function projectWasJustCreated (projectId) {
   return projectId.startsWith('_');
 }
 
-function getSidebarCurrentEle(elem) {
+function getSidebarCurrentEle (elem) {
   let projectId;
   let sidebarRoot;
   let sidebarColorEle;
@@ -140,7 +141,7 @@ function getSidebarCurrentEle(elem) {
   return sidebarCurrentEle;
 }
 
-function getProjectNames(elem) {
+function getProjectNames (elem) {
   // Return a function for timer link to use, in order for projects to be retrieved
   // at the moment the button is clicked (rather than only on load)
   return () => {
@@ -164,3 +165,12 @@ function getProjectNames(elem) {
     return projectNames;
   };
 }
+
+/**
+ * Performance hack for todoist task item re-renders
+ * https://github.com/toggl/toggl-button/issues/1275
+ */
+(function todoistTaskIntegration () {
+  const editor = document.getElementById(EDITOR_ID);
+  editor.addEventListener('mouseover', insertTogglButton);
+})();
--- a/src/scripts/content/todoist.js
+++ b/src/scripts/content/todoist.js
@@ -55,7 +55,8 @@ function insertTogglButton (event) {
       tags: tagsSelector
     });
 
-    container.insertBefore(link, container.lastChild);
+    const button = container.insertBefore(link, container.lastChild);
+    target.addEventListener('mouseleave', () => button.remove(), { once: true });
   }
 }
 
--- a/src/scripts/content/todoist.js
+++ b/src/scripts/content/todoist.js
@@ -53,6 +53,7 @@ function insertTogglButton (event) {
       className: 'todoist',
       description: descriptionSelector,
       projectName: getProjectNames(taskItem),
+      buttonType: 'minimal',
       tags: tagsSelector
     });
 
--- a/src/scripts/content/todoist.js
+++ b/src/scripts/content/todoist.js
@@ -2,15 +2,16 @@
 /* global togglbutton, $ */
 
 const EDITOR_ID = 'editor';
-const TASK_ELEMENT = '.content';
+const TASK_ELEMENT = '.task_item';
 const TOGGL_BUTTON = '.toggl-button';
 
 function insertTogglButton (event) {
   const { target } = event;
-  const shouldInsert = target.matches(TASK_ELEMENT) && !target.querySelector(TOGGL_BUTTON);
+  const taskItem = target.closest(TASK_ELEMENT);
+  const shouldInsert = taskItem && !taskItem.querySelector(TOGGL_BUTTON);
 
   if (shouldInsert) {
-    const container = target.querySelector('.text');
+    const container = taskItem.querySelector('.text');
     const descriptionSelector = () => {
       const clone = container.cloneNode(true);
       let i = 0;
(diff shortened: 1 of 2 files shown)


src/scripts/content/todoist.js
```diff
@@ -1,12 +1,16 @@
...
-  { observe: true },
-  elem => {
-    const container = $('.text', elem);
+const EDITOR_ID = 'editor';
+const TASK_ELEMENT = '.content';
+const TOGGL_BUTTON = '.toggl-button';
 
+function insertTogglButton (event) {
+  const { target } = event;
+  const shouldInsert = target.matches(TASK_ELEMENT) && !target.querySelector(TOGGL_BUTTON);
```
mgsloan: This doesn't work in all cases.  The problem is that the target will report the deepest DOM node that is hovered, but this is only checking for one particular match.  For example, if you put the mouse over the priority indication / checkbox circles on the left and move the mouse through the list, the toggl button does not appear.

I think this should look through the parents and find one that matches `.task`, the class used on the `<li>` for every task.  Perhaps use another variable, like `var task = null;` that gets set if a `.task` is found in the parents.

mgsloan: I also don't think that the `!target.querySelector(TOGGL_BUTTON)` part should be necessary.

I think it might also be good to avoid creating the button if it is already a child of the task element.  Perhaps this isn't necessary, because I'm not seeing duplicated toggl buttons.  However, it seems to me like a fair bit of work is being done whenever `shouldInsert` is `true`, that I think could be avoided if it already exists.

mgsloan:
@shantanuraj Good stuff, I really appreciate you working on this!

I noticed that it doesn't work for all varieties of hovering.  I left a comment on the relevant line of code.

src/scripts/content/todoist.js
shantanuraj: `target.querySelector(TOGGL_BUTTON)` checks just that if the list item already has the toggl button inserted into it. Removing it causes multiple instances to be added to a task row.

shantanuraj:
@mgsloan Ready for another round of testing
Did some basic perf testing most calls to `insertTogglButton` finish execution under a millisecond
The DOM subtree we deal with, is pretty shallow.
Changed the `shouldInsert` calculation to work the task list item instead of the task content.

Let me know if it works better/worse for you

mgsloan:
@shantanuraj Thanks!  Now it the toggl button shows up when hovering other parts of the task.  However, now the button does not disappear when the mouse moves off the task.  I believe the change to the CSS should be reverted.

shantanuraj:
@mgsloan 

This is the behaviour on this branch (with the `visibility: hidden` CSS bits removed)

css-fix

and this is the behaviour if I keep the `visibility: hidden` css

without-css-fix

On hovering the mouse over the task row without the fix you get a phantom increase in height of the row (since the button's there but "hidden") unless you move your cursor over to the content

mgsloan:
@shantanuraj Yeah, that is a bit jarring.  One way to mitigate this problem is to not have the text "Start timer".  Here is some CSS that I think makes it fit in better with Todoist's design:



```css
.toggl-button.todoist {
  font-size: 0px;
  background-size: 18px;
  padding-left: 18px;
  height: 18px;
  position: relative;
  top: -2px;
  left: 2px;
  visibility: hidden;
  background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACYAAAAnCAYAAABjYToLAAAABmJLR0QA/wD/AP+gvaeTAAAACXBIWXMAAAsTAAALEwEAmpwYAAAAB3RJTUUH3ggVDRsAErljWgAAAB1pVFh0Q29tbWVudAAAAAAAQ3JlYXRlZCB3aXRoIEdJTVBkLmUHAAAHm0lEQVRYw7VYX0xU2R3+zrn3zh2GGUYQo51BM0MAI42LsJi2amZDfGhSNxrbh64WTNQlaXja2Cdft0nVNOWtmggojthNY1INSUvrv2xQsiKsLSAS0xRHYEZBYB1ngJm599xfH+TODojei66/l5m598w5v/v9vt/3nXsYbMS5c+dw5MgRAMDFixddhmH4JEmq5ZzvNAzjIwB+IsoHAMbYHIAo53zQMIwewzD6GWOxhoaGeQDo6OhAfX295ZoMNuPUqVOKz+f7nSRJnwEo03U9n3MOACCipZOyV9MahgFFUZJE9D/DML4C8Kf6+nrdznqWibW3t6uKovySiFokScrXNA3vEoqiQAgxR0SfE9GVQ4cOpd82nq908dKlSwCAcDhcoSjKLQB/IaJ3TgoANE0DEeVLkvSVJEk329vbywGgtbV1dYh1dHRUS5LUreu6e3mp3jcYY5BlOZFKpT45fPjwvy0RC4fD5udnnPO+D5GUyUld1z1Op7MvHA7/OnftNyIWDoe3ybJ8T9d1ZbWL5RLfbsiynBFC/KShoeE/b0zs7NmzQbfbPSiEsI0UEaGgoAClpaVERBgdHUUikWB2E2SMQZKkxMLCwkdHjx6NLEns9OnTkCRJzc/Pv0lEO1dTPiKC3+/H7t27AQDXr1+np0+fstUgxxgDY+xOKpXanclkMk1NTeAtLS1oamqCqqp7GGOrTsowjNfKaRgGVjsPY2yXoih7mpqa0NLSAt7Y2PiqCzg/l7uIVei6DkVRUFhYSEKI7HUhBAoLC0mWZei6bns+wzDAGDsPAI2Nja+6sq2t7ZjD4fDaeUoigizLqKmpoQMHDiAYDC5JQAiB0tJSHDx4EDU1NSTLMuzO63A4vG1tbccAgDc3N+c5nc6GTCZj66lcLhd9+umnqKqqYua1lcYBQFVVFduzZw9cLhfZqUYmk4HT6Wxobm7O4x6Px8cYK7MDt8vlon379jGPxwMAiMfjiMVikGU5t/0Ri8UQj8cBAAUFBdi7dy9zuVxksxHKPB6PjwshPhZCuO1Yys6dO6Eor+TtwYMH1NnZiampKSZJ0veKzTkmJydZZ2cnhoeHCQAcDgd27NgBO5YmhHALIT7msiyHrFpbCIFNmzaRz+djADA2Noa+vj7GGAPn/DUOcc7BGMO9e/fY+Pg4AMDv97ONGzcuaZS32FWIc8632YG4trbWJCn19vbCitRmk9y9e5cA0OIczE4jcM63cQAlVoPdbjd5vV4GAKOjo1n+2Il4PM4eP34MAFizZg3cbjfZsLYSzhiz5Jfb/f2QJ0+eLCG7DS9EJJJ1GpiNY1FONyciyWqgw+HIWsz8/DxWazdzc3PZ72bzWKAmyUSUtqHyRESMMQaHw/EatzRNw9TUFBERlushEUFV1ex3K/IvjkvLuq7Pqqq6/m08M58YAHw+H8bHx2FKBGMM09PT6OrqYrkdmdvRPp8v+zuZTFoirOv6LE+lUqNWpUkkEiyVSgEANm/enEVguUSYLye5oaoqKioqAAALCwtIJBLMKrFUKjXKZ2Zm7lvZha7rGBwcpMWXClZdXW3LYoQQqK6uhqIoDAAGBgYs/2cYBmZnZ+/zaDT6DVnohaIoGBkZYS9fviQAqKysZIFAgIQQK2qZyaVgMIjKykpTNujRo0fMqqOJiCYmJr7hc3Nzg0KI51ZaxjlHT09PFu5QKMTq6uqQl5dH6XQamUwGmUwG6XQaeXl5VFdXh1AolOVbT0/PiqVe4YGm5ufnBxkA5cSJE1cDgcAvrLxM13UEg0EKhUIsd5Hp6WmamZkBAKxduxbFxcUst5zd3d0UiUQs0VIUBZFI5O/Hjx/fLwEwksnkd7W1tb+xYRV48eIFGxsbo6KiIuTn57PFXQcrLi5mxcXFzOVyZZOanJykW7duYXJyconRvw2x8+fPfxGLxR5JABCNRkcrKip+6vP5yqzIudg1bGRkhE1PT5PD4YCiKMzkVTqdxrNnz6i3txf9/f1M0zRmVULTIYaGhv55+fLl3wMgE1ujq6vry7Kysh2qqnrsqLnD4UAsFmPmfsxUdE3ToOs6M7c7diOZTCa6urq+BCAAQMqBPVZUVFS8ZcuWn9lR5+XaJYTIqvqbNO1Noaoqbt68+edr1661mzuR3MLTwMBA//r167cGAoHy1byYvE/Isozbt2//o7W19RiArMUsZ+TCw4cP75WXl+/YsGGD70Mnt6iP3545c+ZzTdOiufekFbbQs93d3X3BYHCr3+/f+KGSk2UZ9+/f7zl58uRvNU17uPz+m3r4WX9//x0A6ysrK3+80uHcuwbnHJIk4erVq3+9cOHCF0KI4ZXGSW/xuZnh4eHusbGx70pKSkoLCgoKZVnGuyJobsWj0ejj1tbWP964ceMPQoix9znqVFwuV+3WrVsP7t+//1eBQOBH5jugnTAlIxKJxK5cufK3oaGhS/Pz898C0H6QM1gAbgDbtm/f/vNdu3Z9UlJSUu71et2yLDslSZKXoa3rup6Kx+PJiYmJ/965c+frvr6+fwEYAJD8Qc5gVwIBwFqv1xtYt25dUFVVv9PpXKuqah4ApNPphVQqNZNOp6PPnz9/HI/HIwBmAGRWs8j/AVpCzgNn8cBUAAAAAElFTkSuQmCC);
}

.task_item:hover .toggl-button {
  visibility: visible;
}
```

shantanuraj:
@mgsloan 

- Have switched to the `minimal` button
- Incorporated some of your CSS

I am apprehensive about including the visibility hidden bit in this as it leads to the phantom height increase of the task rows, particularly in the "Overdue" tasks list it works fine in the "Today" list.
For that case removing the button seems like the better solution (restores the row's height on removing focus from the row)

## visibility-hidden CSS
hidden

## remove button on mouse leave
remove

I hope that works for you

mgsloan:
I like the use of the minimal button!  Makes the CSS simpler.

Yeah, the content shifting isn't good.  This CSS resolves the problem:

```
.toggl-button.todoist {
  padding-left: 20px;
  margin-bottom: -6px;
  visibility: hidden;
}

.task_item:hover .toggl-button {
  visibility: visible;
}
```

Specifically, the toggl button was too large vertically, and so `margin-bottom: -6px` reduces the layout height.  I also made the following changes:

* `visibility: hidden` when unhovered.  It looks weird to leave a trail of buttons of just the things that have been hovered.

* hover css is on `task_item` instead of `task_content_item`.  Otherwise the hover behavior is inconsistent with todoist's icon and button insertion.

I agree that it would be ideal to avoid task height.  Perhaps there is a decent solution that involves the mutation observer, perhaps involving debouncing mutation handling..

However, with the minimal button, the task height does not often change.  In my "ideas" super-project it only affects 6 out of 300 items.  I took a look at one of these cases, and removing one character from the task makes the issue go away.  Adding two characters makes the issue go away.  So, in other words, this only affects tasks where the last line has a specific length, or one character more than the length.

mgsloan:
I've figured out one good reason to eagerly populate the toggl buttons rather than waiting for focus.  Specifically, I'd like it if `.toggl-button.active` (when a task is being tracked) was always visible.

I have a specific proposal for how to do eager population:

* Have a mutation observer like the one, which sets a boolean flag, perhaps `tasksMutated` to `true` indicating that the task list may have changed. 

* When a focus occurs, immediately populate the toggl button for that task.  Then, defer an action to populate the rest of the tasks.

I implemented this, but ran into the same slowdown.  My theory was wrong, I think that creation of many toggl buttons is inherently inefficient.  Specifically, it takes about 700ms to do the initial `createTimerLink`.  However, *every single invocation* sends a chrome message to the background page asking if it's the current task (`currententry`).  Every single handler for this response then re-queries the description from the DOM.  On my computer, with a 4ghz quadcore xeon, handlers take 6 seconds for 300 tasks.

The `currententry` message should only be sent once.  This should cut down 6 seconds of the browser going out to lunch.  The 700ms to create all the timer links is still non-ideal.  I think the best solution is to create the DOM nodes for the timer links, but defer all the computation until focus.

mgsloan:
I am working on a promising approach which allows for super fast creation of toggl links.  This allows deferral of all of the costly bits of timer link initialization

mgsloan:
Sorry for suggesting the focus stuff!  Turns out to probably be the wrong direction.  I've opened some PRs which I think address the issue a bit more systemically:

* https://github.com/toggl/toggl-button/pull/1281 which should greatly improve performance of toggl-button across many of the integrations.

* https://github.com/toggl/toggl-button/pull/1282 for using the minimal button with the todoist integration.

I still think there are some performance issues here.  Specifically, it takes about 700ms to render 300 toggl buttons, with the todoist integration, which is a quite noticeable lag.

I did an experiment where very little computation is done when creating the button (no figuring out of description text or project info) - deferring this info until the hover over the task.  This worked pretty darn well.  Might make a PR that does something like that in the future after the above ones are merged.

shantanuraj:
Closing in favour of #1281 and #1282

PR #1281, opened 2019-02-09
Improve performance for pages with many toggl buttons + cleanups + todoist performance improvements

mgsloan:

This PR makes some changes to `common.js`, along with minimal changes to `todoist.js` in order to mostly resolve the performance issues described in https://github.com/toggl/toggl-button/issues/1275

* Extracts out a function called `queryAndUpdateTimerLink` (better name suggestions appreciated).

* `queryAndUpdateTimerLink` is now run at the end of `renderTo` instead of at the end of `createTimerLink`. 
 
    - Here's why:  Code similar to `queryAndUpdateTimerLink` was being run after every single `createTimerLink` invocation.  This was tremendously inefficient, because it sends a message to the toggl-button background page asking it to say what the currently tracked task is.  The handler for every response to these calls would then scan every single link and modify it via activate / deactivate.  This is also discussed some [here](https://github.com/toggl/toggl-button/pull/1277#issuecomment-462013597).

    - This is still not really ideal, and probably runs `queryAndUpdateTimerLink` waaay more than should be necessary.

* Adds optional support for debouncing mutation observer results.  This is helpful if many mutations are expected to happen in rapid succession.  It also effectively defers rendering of toggl buttons until the page isn't as busy.

* Adds optional support for a specific target for the mutation observation, making it more efficient.

* Cleans up some code in `common.js`.

* Modifies the todoist integration to use mutation observer debouncing and targeting.

Note that it may be confusing to review this PR via "files changed", please take a look at individual commits.

Changes have been tested using the todoist integration in chrome.

* https://github.com/toggl/toggl-button/issues/1275

* https://github.com/toggl/toggl-button/pull/1277#issuecomment-462013597

### DIFF
--- a/src/scripts/common.js
+++ b/src/scripts/common.js
@@ -433,6 +433,12 @@ window.togglbutton = {
   },
 
   createTimerLink: function (params) {
+    const link = togglbutton.createTimerLinkSkipQuery(params);
+    togglbutton.queryAndUpdateTimerLink();
+    return link;
+  },
+
+  createTimerLinkSkipQuery: function (params) {
     let link = createLink('toggl-button');
 
     const project = invokeIfFunction(params.projectName);
@@ -511,26 +517,14 @@ window.togglbutton = {
     // Add created link to links array
     togglbutton.links.push({ params: params, link: link });
 
-    // new button created - set state
-    chrome.runtime.sendMessage({ type: 'currentEntry' }, function (response) {
-      let currentEntry; let i;
-      if (response.success) {
-        currentEntry = response.currentEntry;
-        for (i = 0; i < togglbutton.links.length; i++) {
-          link = togglbutton.links[i].link;
-          if (
-            invokeIfFunction(togglbutton.links[i].params.description) ===
-            currentEntry.description
-          ) {
-            activate();
-          } else {
-            deactivate();
-          }
-        }
-      }
-    });
-
     return link;
+  },
+
+  // Query active timer entry, and set it to active.
+  queryAndUpdateTimerLink: function () {
+    chrome.runtime.sendMessage({ type: 'currentEntry' }, function (response) {
+      togglbutton.updateTimerLink(response.currentEntry);
+    });
   },
 
   // If "entry" is passed, make button active; otherwise inactive.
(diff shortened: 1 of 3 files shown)


src/scripts/common.js
```diff
@@ -512,66 +502,58 @@ window.togglbutton = {
...
-
-    let link;
-
-    let i;
-
-    let minimal;
+    if (entry) {
+      for (const current of togglbutton.links) {
+        if (invokeIfFunction(current.params.description) !== entry.description) {
+          togglbutton.deactivateTimerLink(current.link);
```
shantanuraj: Is this required, since `activateTimerLink` at L524 will deactivate all other links anyway?

shantanuraj: If not, we could make it simpler by replacing L520-L529 with

```javascript
const current = togglbutton.links.find(button => invokeIfFunction(button.params.description) === entry.description);
togglbutton.activateTimerLink(current.link);
```

src/scripts/common.js
```diff
@@ -118,20 +118,32 @@ window.togglbutton = {
...
-                return;
+              if (mutationSelector) {
+                const matches = mutations.filter(function (mutation) {
+                  return mutation.target.matches(mutationSelector);
+                });
+                if (!matches.length) {
+                  return;
+                }
+              }
+              if (opts.debounceObserve > 0) {
```
shantanuraj: `debounceInterval` might be a better choice for the option

src/scripts/common.js
tcrammond: Adding in here that we've seen at least one integration in testing that now gets stuck in an infinite re-rendering loop (Atlassian Jira Cloud, issue page) when the button becomes "active". Wondering if it's related to the above *(I have not debugged it fully myself at this time, sorry)*.

src/scripts/common.js
```diff
@@ -512,66 +502,58 @@ window.togglbutton = {
     // Add created link to links array
     togglbutton.links.push({ params: params, link: link });
 
-    // new button created - set state
+    return link;
+  },
+
+  // Query active timer entry, and set it to active. If a callback is
+  // provided as an argument, it will be invoked before updating the
```
tcrammond: It seems that a callback isn't an argument here in `queryAndUpdateTimerLink`, I assume it wasn't needed after all?

src/scripts/common.js
mgsloan: :+1: Modified it to avoid the redundant deactivations and use `find`

mgsloan:
@tcrammond Thanks, I appreciate your review!

PR #1308, opened 2019-02-15
Refactor "link activation" to query DOM instead of storing list of links

tcrammond:
*Merging to feature branch*.

Cleans up link activation/deactivation further.

The `togglbutton.links` list was in fact a "memory leak" of detached DOM elements and would infinitely increase in size over time. I did not see it used anywhere else.

It also caused buggy functionality with "activating" the active timer link on the page, due to old link references never being removed.

Also a bonus fix for stopped links appearing green when they should not. 

Second commit fixes "infinite re-render". Upon activation, the button class list would be mutated too much and cause re-renders to happen in integrations that did not have mutation selectors. In future, we should seek to completely construct the button before adding it to the DOM. There is an opportunity to *always* include `*:not(.toggl-button)` in mutation selectors by default in future, but I did not do that here.

Related to #1275.

Recommended to test that Trello or Todoist perform as expected. I developed against Trello.

Links should activate/deactivate correctly. So if you have a timer running, and you open the related Trello card for example, the link will show as `Stop Timer` because it's already running, no matter how many times you open and close the card.

### DIFF
--- a/src/scripts/common.js
+++ b/src/scripts/common.js
@@ -95,7 +95,6 @@ window.togglbutton = {
   $billable: null,
   isStarted: false,
   element: null,
-  links: [],
   serviceName: '',
   projectBlurTrigger: null,
   taskBlurTrigger: null,
@@ -499,9 +498,6 @@ window.togglbutton = {
       return false;
     });
 
-    // Add created link to links array
-    togglbutton.links.push({ params: params, link: link });
-
     return link;
   },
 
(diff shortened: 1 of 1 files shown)


src/scripts/common.js
```diff
@@ -514,10 +510,16 @@ window.togglbutton = {
...
   updateTimerLink: function (entry) {
-    const current = togglbutton.links.find(
-      button => invokeIfFunction(button.params.description) === entry.description);
+    if (!entry) {
+      togglbutton.deactivateAllTimerLinks();
+      return;
+    }
+
+    const current = Array.from(document.querySelectorAll('.toggl-button:not(.toggl-button-edit-form-button)'))
+      .find(button => button.title.indexOf(entry.description) !== -1);
```
tcrammond: Note: the `title` attribute is set to `description + (project if project)` at insertion time, this seems to be the cheapest way to check if it's the correct entry. Shouldn't be necessary to run the description/project functions again and do loads of look ups (that would get quite intense with many links in a render-heavy service). 

There is perhaps a chance of an edge case where wrong entry is highlighted because it still matches in `indexOf`, but it seemed uncommon enough to be an acceptable compromise. Open to changing it.

Issue #1334, opened 2019-03-04
Q: How Customizable is the Database? (from the dev-end)


BeksOmega:
### Problem Statement

I'm working on an integration with workflowy.com (who I am _not associated with_) and I was wondering if the functionality I want to add would be possible with your current database setup.

workflowy is basically a tree structure, with parents, children, and siblings:

![structure](https://user-images.githubusercontent.com/25440652/53742367-54f11380-3e4d-11e9-91f6-ae19de9699a8.jpg)

I want to use this to semi-intelligently predict what project the user will assign to a given node. It would work by assuming people group tasks in the same project together.

For example, say you assigned the RedProject to node A1. The extension would then associate that project with nodes A1, A, and Top Node.

![assignred](https://user-images.githubusercontent.com/25440652/53742441-77832c80-3e4d-11e9-9fa8-054b587818c7.jpg)

Then if you went to start a timer for A1b It would automatically populate the project slot with RedProject, because that is the closest assigned project (the great thing about this method is it would only have to traverse parents in the tree, not siblings or children).

![detectred](https://user-images.githubusercontent.com/25440652/53742503-9aaddc00-3e4d-11e9-8f40-8d111c0d949c.jpg)

Then say you want to start a timer for A2a. It automaticaly populates with RedProject,

![detectred2](https://user-images.githubusercontent.com/25440652/53742766-435c3b80-3e4e-11e9-883d-8440a812cb3f.jpg)

but then you manually set it to BlueProject. The extension then climbs the tree, associating parent nodes with the new project until it reaches one that has already been assigned.

![assignblue](https://user-images.githubusercontent.com/25440652/53742571-cd57d480-3e4d-11e9-93d6-3cc2ce1c7b5c.jpg)

The great thing is that as long as I can manually assign simple key-value pairs this should work, since each node has a custom id that is available through the DOM. I would just assign (nodeID -> projectID) and do the traversal inside the projectSelector.

tcrammond:
Hi @BeksOmega, the simplest answer is: no, you can't assign any key value pairs or anything like that with projects in Toggl Button. We do have long-term wishes to support better project selection options.

Something that may inspire you is the Todoist integration. When starting a timer, it climbs the nested tree of Todoist projects and tries to find the closest leaf that matches an existing project in Toggl. You could achieve something similar to this with the Workflowy nodes (just watch out for performance implications in a big document). If you decide to tackle it, please make a separate PR for it after your first one is released 😄

PR #1335, opened 2019-03-06
feat: Added workflowy.com integration

BeksOmega:

Adds the toggl button to [workflowy.com](https://workflowy.com/) in two places.

In the hover popup:
hover

In the more options menu:
options

### Finding the Hover Button:
1. Navigated to [workflowy.com](https://workflowy.com/).
2. Created a bullet.
3. Hovered over the bullet (the circle part).
4. Observed how the toggl button appeared. (pass)

### Finding the More Options Button:
1. Navigated to [workflowy.com](https://workflowy.com/).
2. Created a bullet.
3. Clicked on the bullet (the circle part). This makes it so that that bullet is now the top bullet.
4. Clicked the more options menu on the right.
5. Observed how the toggl button appeared. (pass).

### Does it Populate the Toggl Button Fields Correctly?

The field selectors traverse the DOM to grab the description & tags, and because of how workflowy structures its DOM this can get tricky.

Example Bullet:
bullet

Example DOM for Bullet:
dom

Button Popup: (pass)
buttonpopup

### Does it Change Themes Correctly?
Default: (pass)
default

Vintage: (pass)
vintage

Hacker: (Workflowy Pro only) (pass)
hacker

### Gif of the Button in Action:
extensionexample

Related to #1334

### DIFF
--- a/src/scripts/common.js
+++ b/src/scripts/common.js
@@ -172,11 +172,16 @@ window.togglbutton = {
     let len;
 
     const elems = document.querySelectorAll(selector);
+    if (!elems.length) {
+      return;
+    }
+
     for (i = 0, len = elems.length; i < len; i += 1) {
       elems[i].classList.add('toggl');
     }
 
-    // Catch content errors here as well as render() in case of async rendering (MutationObserver)
+    // Catch content errors here as well as render() in case of async rendering
+    // (MutationObserver)
     try {
       for (i = 0, len = elems.length; i < len; i += 1) {
         renderer(elems[i]);
@@ -210,7 +215,8 @@ window.togglbutton = {
 
     const description = togglbutton.mainDescription.toLowerCase();
 
-    const projectId = togglbutton.findProjectIdByName(togglbutton.currentProject);
+    const projectId = togglbutton.findProjectIdByName(
+      togglbutton.currentProject);
 
     if (togglbutton.entries) {
       togglbutton.entries.forEach(function (entry) {
@@ -300,7 +306,8 @@ window.togglbutton = {
 
     const elemRect = togglbutton.element.getBoundingClientRect();
     editForm = $('#toggl-button-edit-form');
-    const position = togglbutton.topPosition(elemRect, editFormWidth, editFormHeight);
+    const position = togglbutton.topPosition(
+      elemRect, editFormWidth, editFormHeight);
 
     if (editForm !== null) {
       togglButtonDescription = $('#toggl-button-description');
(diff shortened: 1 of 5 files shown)


src/scripts/common.js
```diff
@@ -163,11 +164,16 @@ window.togglbutton = {
     let len;
 
     const elems = document.querySelectorAll(selector);
+    if (!elems.length) {
+      return;
+    }
+
```
BeksOmega: The above is the only actual change in this file, the other things were just to get it to pass lint.

BeksOmega:
Good to go.

src/styles/style.css
```diff
@@ -1587,3 +1587,20 @@ label > .toggl-button.outlook {
 .toggl-button.dropbox-paper {
   vertical-align: middle;
 }
+
+/****** WORKFLOWY ******/
+.workflowy-toggl-wrapper {
+  display: flex !important;
+}
+
+.toggl-button.workflowy {
```
tcrammond: This is a global problem, but in the meantime I think we'll need to add `:not(.toggl-button-edit-form-button)` to stop some of these styles leaking to the popup that appears after you click a button.

tcrammond: It would also be a bonus if the styles in the menu could be tweaked so that it fits more naturally in the Workflowy UI. At the moment, the text and icon are noticeably bigger.

Screen Shot 2019-03-25 at 16 49 03

BeksOmega:
- Fixed styles leaking to the popup. The one thing I noticed that's still leaking though is the color, but I wasn't sure what you wanted the default color to be.
- Tweaked the button style so that the text size will (hopefully) match the size of the other text. I wasn't able to replicate this problem on windows (as evidenced by the pictures in the OP) so you'll have to tell me if my changes helped or not (hopefully they did!).
- Tweaked the button style so the icon is smaller.

Currently it is the same height as the text (14px):
ButtonStyle_14px

Originally it was ~20px:
ButtonStyle_Auto

- I was unable to reproduce the hover menu problem. If I hover over the bullet that started the timer it says "stop timer" and if I hover over a different bullet it says "start timer". If this is not the intendend behavior I'm still happy to fix it =)

StopStartExample

- I don't believe the tags problem is related to my selector. If you add the below code at ~ln 118 in workflowy.js
```
for (var i = 0; i < tagsArray.length; i++) {
    console.log(tagsArray[i]);
  }
```
you can see that the tagsArray is being filled correctly, they're just not being displayed. If you look at the google keep integration you can see that the tags are also not being populated there. I could very well be missing something though.

BeksOmega:
> re. tags: It was my mistake - I didn't have "automatic tagging" setting enabled.

Ah thank you! Looks like I should turn that on too hehe.

PR #1352, opened 2019-03-20
1289/standard button style

### LINKED ISSUE #1289: Standardise integration button presentation
The way in which integration buttons are presented and behave can vary between services.

Let's determine and write up a "standard", and update at least the most popular integrations to follow this.

Notes:

* Some buttons have transparency/opacity animations, or even colour changes via CSS (mainly for dark modes).
* Some buttons in lists-of-tasks show only on hover, some show all the time and look messy.
* Often the button CSS interferes with the timer popup, because it has the same base class name.

Tasks:

* Rules in CONTRIBUTING and PR template html comment
* Refactor top ~20 integrations to follow rules

Suggestion action based on implied standards we kind of already had, can/will be updated:

* **Standalone integration buttons**
  * Should appear as `[Red icon] Start timer`.
  * `minimal` mode may be used if the button is in a tight spot, such as a toolbar.
  * If it would fit nicely in a UI component of the integration (e.g. Trello card button), do so.
* **List integration buttons (e.g. Todoist)**
  * Always use `minimal` mode.
  * Buttons must only show on hover, in UI context menus, or both.
  * Non-active buttons must use the grey icon.
  * Active timer button must use the red icon
  * Performance must have been checked. We don't want to degrade the performance of the integration.
* **All buttons**
  * The style of other parts of the integration's UI must not be affected.
  * The button must not cause the integration's UI to flow badly or change in some other way.
  * The popup shown after clicking the button should have standard styles (often, in the past, style has leaked onto the 'Stop timer' button in the popup.)
  * The text used as timer description should be based on the most general use case for users of the integration service. E.g. if ticket ID is an important piece of information, it should be included. If it isn't important information (such as it's never actually shown in the UI, only in URLs) it should **not** be included.

tcrammond:

* Adds 'style guide' for integration buttons
* Trello: Improve/fix layout of all buttons
* Zendesk: Improve/fix layout
* Google mail: Tweak button style
* Github: Fix wrong issue number being picked up in Project Board button
* Google docs: minor style fixes

I checked top ~20 integrations on the list.

Closes #1289

### DIFF
--- a/src/scripts/content/trello.js
+++ b/src/scripts/content/trello.js
@@ -50,14 +50,14 @@ togglbutton.render(
     };
 
     const link = togglbutton.createTimerLink({
-      className: 'trello',
+      className: 'trello-list',
       buttonType: 'minimal',
       projectName: getProject,
       description: titleElem.textContent + ' - ' + taskElem.textContent
     });
 
-    link.classList.add('checklist-item-button');
-    elem.parentNode.appendChild(link);
+    link.classList.add('checklist-item-menu');
+    elem.querySelector('.checklist-item-menu-wrapper').appendChild(link);
   },
   '.checklist-items-list, .window-wrapper'
 );
--- a/src/scripts/content/trello.js
+++ b/src/scripts/content/trello.js
@@ -25,6 +25,12 @@ togglbutton.render(
       projectName: getProject
     });
 
+    // Pass through click on Trello button to the timer link
+    container.addEventListener('click', (e) => {
+      e.preventDefault();
+      link.click();
+    });
+
     container.appendChild(link);
     descriptionElem.parentNode.insertBefore(container, descriptionElem);
   },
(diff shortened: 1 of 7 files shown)


src/scripts/content/trello.js
```diff
@@ -25,6 +25,12 @@ togglbutton.render(
       projectName: getProject
     });
 
+    // Pass through click on Trello button to the timer link
+    container.addEventListener('click', (e) => {
```
shantanuraj: what a lovely hack 😄

src/scripts/content/zendesk.js
```diff
@@ -27,6 +26,5 @@ togglbutton.render('.pane_header:not(.toggl)', { observe: true }, function (
     projectName: projectName && projectName.split(' - ').shift()
   });
 
-  divTag.appendChild(link);
-  elem.insertBefore(divTag, elem.querySelector('.btn-group'));
+  elem.insertBefore(link, elem.querySelector('.btn-group'));
```
shantanuraj: There's some problems with the TE popup and Zendesk

- Link always says "Start timer" even if timer is already running
- The title attribute is `undefined`

See link
https://streamja.com/2l4z

PR #1354, opened 2019-03-21
fix: Fix errors on workdayEnd when $curEntry is null

tcrammond:

Resolve https://app.bugsnag.com/toggl-dot-com/toggl-button/errors/5c90c3e73728f4001bb193b7?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[release.seen_in][0][type]=eq&filters[release.seen_in][0][value]=latest&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[event.class][0][type]=ne&filters[event.class][0][value]=Content

Add a defensive check. This seems to have come about because of async functions and the various ways this can be triggered. Not looking to refactor the flow at this time.

Should look like this error can not happen anymore even if it wanted to.

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -1435,6 +1435,10 @@ window.TogglButton = {
   },
 
   checkWorkDayEnd: async function () {
+    if (!TogglButton.$curEntry) {
+      return;
+    }
+
     const stopAtDayEnd = await db.get('stopAtDayEnd');
     const hasWorkdayEnded = await TogglButton.workdayEnded();
     if (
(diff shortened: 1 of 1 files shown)


src/scripts/background.js
```diff
@@ -1435,6 +1435,10 @@ window.TogglButton = {
   },
 
   checkWorkDayEnd: async function () {
+    if (!TogglButton.$curEntry) {
+      return;
+    }
+
     const stopAtDayEnd = await db.get('stopAtDayEnd');
     const hasWorkdayEnded = await TogglButton.workdayEnded();
```
Phenome: This would return false from `workdayEnded` if `!TogglButton.$curEntry`, anyway. But also doesn't harm to also check here :)

Issue #1357, opened 2019-03-22
Unable to dismiss the "Welcome to Toggl Button!" popup in settings


devsuitup:
Please follow this guide when creating a new issue. This speeds up the process of replicating the issue and finding a solution.

### Environment details

* Browser : Firefox 65.0.2 (64 bits)
* Operation system: Windows 10
* Extension version: 1.23.4

### Expected behavior
Being able to dismiss the popup to access settings

### Actual behavior
The popup doesn't close.

### Steps to reproduce the behavior
New installation in Firefox, then going to settings

### Additional details
In the browser console :
TypeError: NetworkError when attempting to fetch resource.
URL de la ressource : moz-extension://4b89327d-5531-4826-8157-9e32d28ef34d/scripts/settings.js
URL du lien source : settings.js.map[En savoir plus]

* screenshots or screencapture
![image](https://user-images.githubusercontent.com/2380987/54820225-af3c0200-4c9e-11e9-839c-e5cdbdd73c1f.png)

flexchar:
Same on Firefox 67 on Win 10 / macOS Mojave

tcrammond:
This issue should be resolved in versions `1.24.0` onwards. Sorry for the inconvenience, please let us know if anything else comes up with the settings.

PR #1358, opened 2019-03-22
Ensure default settings are loaded in case of invalid value

tcrammond:

Ensure default settings are loaded in case of invalid value. Horrible bugs aside, this is valuable so that user can recover seamlessly if anything does go wrong with the stored settings.

`async`-ness of `load()` is removed because it was not used correctly (😬) and can definitely cause severe settings problems in some browsers, it seems.

First time I loaded Toggl Button into Firefox **Developer** Edition, all of my settings became `null` 😱 . The root cause is probably the incorrect `async` usage - each value was not retrieved before it was attempted to be saved again in `load`, so it was saved as `null`. *(This re-saving is on the cards to remove, but this fix is still needed in the meantime).*

Please check all of the browsers, including Firefox Developer Edition. Settings should load flawlessly on reloads, etc. Inspect the background console.

Should fix symptoms of #1357.

### DIFF
--- a/src/scripts/lib/db.js
+++ b/src/scripts/lib/db.js
@@ -24,18 +24,14 @@ const DEFAULT_SETTINGS = {
   dayEndTime: '17:00',
   defaultProject: 0,
   rememberProjectPer: 'false',
-  enableAutoTagging: false
-};
+  enableAutoTagging: false,
 
-// core settings: key, default value
-const CORE_SETTINGS = {
   'dont-show-permissions': false,
   'show-permissions-info': 0,
   'settings-active-tab': 0,
   sendErrorReports: true,
   sendUsageStatistics: true
 };
-
 const transformLegacyValue = (value) => {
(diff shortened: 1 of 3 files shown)


src/scripts/lib/db.js
```diff
@@ -153,14 +149,27 @@ export default class Db {
     this.set(this.togglButton.$user.id + '-defaultProjects', {});
   }
 
-  get (key, defaultValue) {
-    const hasDefaultValue = typeof defaultValue !== 'undefined';
-    return browser.storage.sync.get(hasDefaultValue ? { [key]: defaultValue } : key)
+  get (key) {
+    const hasDefaultValue = typeof DEFAULT_SETTINGS[key] !== 'undefined';
+    const defaultValue = DEFAULT_SETTINGS[key];
```
shantanuraj: Ultra-minor nitpick but you could swap L153 and L154 and use `defaultValue` in the `typeof` check

PR #1363, opened 2019-03-27
Fix wordpress integration

### LINKED ISSUE #1362: Wordpress integration not working
* **OS version**: Linux Mint 19.1
* **Browser version**: Chrome 73.0.3683.75 64 bit
* **Extension version**: 1.24.1

### 🐛 Describe the bug
Tried the integration on WP 4.9.8 and 5.1.1 and couldn't find the button.
### Expected behaviour
Screenshot from the wiki page looks like it's from an old WP version. Button should be present on the `new post` page
image
#### Button missing in WP 5.1.1
image
#### Button missing in WP 4.9.8
image
### Steps to reproduce
* Configure your custom Wordpress domain on the Toggl Button settings
* Verify that the button is missing

Phenome:

Button wasn't appearing on WP 4.8.3, 4.9.8 and 5.1.1
Fixes Wordpress integration for
#### WP 4.x.x
image

#### and WP 5.1.x
image

* I tried on a Wordpress 5.1.1 that had two types of post forms. The new, 5.x.x, and the previous one. Can verify both codes from a single WP installation. I also tested on an actual 4.9.8 installation.

Fixes #1362

### DIFF
--- a/src/scripts/content/wordpress.js
+++ b/src/scripts/content/wordpress.js
@@ -1,22 +1,34 @@
 'use strict';
 
-togglbutton.render('.editor__header:not(.toggl)', { observe: true }, function (
-  elem
-) {
-  const tabs = $('.editor__switch-mode', elem);
-
+// WP 4.9
+togglbutton.render('#poststuff:not(.toggl)', { observe: true }, function (elem) {
+  const heading = document.querySelector('.wp-heading-inline');
   const description = function () {
-    return document.querySelector('.editor-title__input').value;
+    return elem.querySelector('#title').value;
   };
-
-  if ($('.toggl-button')) {
-    return;
-  }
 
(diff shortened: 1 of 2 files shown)


Phenome:
I tried this on a new blog on Wordpress.com and could repro the error. Added a check so it doesn't throw.
But when editing the site through wordpress.com, it uses `iframe` s, and I can't access it to add the button because of cross-origin restrictions.

i.e. I created a `thisbewordpresstest.wordpress.com` and when editing it through wordpress.com, it uses an `iframe`. I'm fairly confident that the most common use case here would be to enable the integration to a custom domain running Wordpress, so when I added the `thisbewordpresstest.wordpress.com` domain and accessed it directly, it worked.

I'm not sure if there's a good way around the `iframe` limitation.

Issue #1290, opened 2019-02-11
Toolbar popup: best effort improvements


tcrammond:
Until we work on a dedicated re-design later, we'll do a best effort clean up and improvement of layout and appearance, possibly lifting some designed **styles** from the web app.

The goal is to improve first impressions and perceived quality.

![screenshot2019-01-31at12 14 17](https://user-images.githubusercontent.com/6432028/52565071-ad5c6480-2dfd-11e9-8320-c6faed5a8561.png)

**Notes**

* We can probably increase the width by 50% or so, to give time entries room to breathe.
* Style of text, projects etc. and buttons
* Layout / styles of time entries list
* Shift the layout around - top half contains relatively unimportant things and unclear buttons.
* Dig up old issues from Indrek, or with user suggestions we've received over time for reference.

**Vague plans:**
- Increase the width a little
- Remove 'continue latest' - this is performed from TEs.
- 'Start new' can probably be a timer button or something (like the webapp), moved alongside the heading of the time entries list.
- Move settings, sync and log out to the bottom. Stop relying on unclear icons. Should we keep the toggl app link? (It should change appearance if so).
- New time entries list layout.
  - TBD, but can leverage webapp *styles* for the icons, text, etc. in whatever the new layout is.
  - (?) Is it viable to pull in the `preact` experiment branch for at least generating HTML  for the new list? *(or whole popup?)*

tcrammond:
Released #1365 🎉 🎉

Issue #1288, opened 2019-02-11
Login: Public web standalone log in page


tcrammond:
Currently the login occurs in the "toolbar pop up", and we have to conditionally render various states in there, including the no-workspace-state. Some users have trouble logging-in via passwords-managers mostly on Firefox.

We will ~move it to a standalone page~ create a new page on public web for Toggl Button log in.

- Remove all login-related views in the toolbar popup
  - Instead, direct to the new button login page when clicking the toolbar button when login is required.
- No workspace state: would it be best if we send people to webapp to resolve this? The alternative is we maintain the existing toolbar popup view.
- Ensure the button still indicates when user needs to log in, any other similar states.

Firefly page:
- Base off existing login page *(check for not needed conditions, states, API calls etc)*
- New copy
- Relevant other useful links - store links? Main toggl button page? Looking for web app instead?
- Login
  - Logs in and redirects to **Toggl Button extension landing page**

Extension landing page:
- Some text saying you're logged in and to start tracking
- Useful help links and settings link etc.

shantanuraj:
I'd say we should explore the possibility of using public web to simplify our login process

This should help us deal with google login and other subsequent changes to the login flow later on (eg. if we add 2FA support)

See trello's login flow below

#### Login page
![image](https://user-images.githubusercontent.com/1716853/52998213-bc1fc880-3448-11e9-9226-6a602ae806d4.png)
#### Oauth login
![image](https://user-images.githubusercontent.com/1716853/52998231-ccd03e80-3448-11e9-8294-5035a39d4906.png)
#### Extension Landing on login
![image](https://user-images.githubusercontent.com/1716853/52998286-ef625780-3448-11e9-945f-f821f4371c17.png)

shantanuraj:
Closed via #1367

Issue #1291, opened 2019-02-11
Timer popup: best effort improvements


tcrammond:
Until we work on a dedicated re-design later, we'll do a best effort clean up and improvement of layout and appearance, possibly lifting some designed **styles** from the web app.

![screenshot2019-01-31at12 15 24](https://user-images.githubusercontent.com/6432028/52565198-06c49380-2dfe-11e9-9800-4d46b5e3375a.png)

Toolbar popup issue here: #1290.

The goal is to improve first impressions and perceived quality.

**Notes**

* Better buttons, icons, colours
* Improve UX of popdowns (Tags popdown is particularly olden)

shantanuraj:
Closed via #1386

PR #1365, opened 2019-03-28
Toolbar popup revamp

tcrammond:

Revamps the "toolbar popup" accessible from the browser toolbar.

I'm sorry commits are a bit intertwined and it's a relatively big PR - these refactors really couldn't be compartmentalised very well. We'll get there eventually.

Feature parity should be maintained, with some extra bonuses. These are the main things it is capable of now:

• (new) Start timer with a description in the timer bar
* Running timer can be clicked to be edited
• Header has icon link to webapp and icon link to settings page. “Log out” and “Sync data” have been moved to a new tab in settings “My Account”.
• Shows today’s and weekly total above the list _(based on the TEs it knows about)_ as in original version
• (new) Grouped entries! Same grouping as webapp. *Cannot* be expanded but has a tooltip `5 entries since hh:mm`.
• Entry layout is now pretty much the same as mobile/desktop, kind of a hybrid between them all.
• (new) Added day headers, marking each day between entries in the list as in other clients
• (new) List is no longer limited to `max. 5 unique TEs`. It is now limited to `max. 15 TE groups` (TEs are still fetched from `/me?with_related_data`, not sure if it will even get that much data)
• (new) At the bottom of the list there is a `See more on toggl.com` button, in lieu of the `Load More` you see in other clients.

*N.B.* The log in form and “edit new entry” form remain as-is and have separate issues for being revamped.

Bigger missing features that button has never had like editing TEs, expanded groups, tasks/clients, datetime format preferences were out of scope for now.

#### Old pop up:

Screen Shot 2019-03-29 at 15 46 43

### New pop up:

Screen Shot 2019-03-29 at 15 33 53

##### Login form (temporary, being removed in #1288)

Screen Shot 2019-03-29 at 16 17 07

##### List scrolled down:

Screen Shot 2019-03-29 at 16 14 03

* Test the popup itself, of course. Starting/stopping/continuing TEs.
* Settings should be played with - there can be knock on effects from browser messaging changes.
* Integrations (such as Trello, Todoist) should be played with - there can be knock on effects from browser messaging changes.

#1290 in Phoenix project.


tcrammond:
> We should update the screenshots on the store listing as well

Yeah, i'll coordinate with support about the KB as well 👍

dooart:
Uhh... since this was not merged yet I'll not open the aforementioned issue yet.

It's just a small bug: if you stop a running entry and immediately start typing a new description in the input, whatever you typed will soon disappear. See it happening [https://toggl.slack.com/files/U0GJ101MJ/FH6S20KMX/button_2.mp4](here).

shantanuraj:
There are potentially some bugs in updating the `TimeEntriesList` component

Just started a TE notice the difference between webapp and button (the "today" total seems to be fine though)

### webapp
image

### button
image

rylek90:
Can we make font in Project popup a little bit more visible? 🙈 
image

Also, when I start new TE and have to select Projects/Tags UI is quite confusing. It's impossible to hide popup by clicking outside of it (for me only clicking on X works)

tcrammond:
@rylek90 these are valid points and fortunately they will receive love and care in #1291 :smile:

src/scripts/@toggl/style/lib/text.js
```diff
@@ -0,0 +1,36 @@
...
+
+import * as color from './color';
+
+export const font = {
+  arial: `Arial, sans-serif`,
+  openSans: `"Open Sans", Arial, sans-serif`,
+  roboto: `Roboto, Helvetica, sans-serif`
+};
+
+export const [light, normal, semibold, bold] = [300, 400, 500, 700];
```
rylek90: 😍 beautiful

src/scripts/components/Timer.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+function RunningTimer(props: { entry: TimeEntry }) {
+  const { entry } = props;
+  const editEntry = (e) => {
+    e.preventDefault();
+    (window as any).PopUp.updateEditForm((window as any).PopUp.$editView);
+  };
+  const stopTimer = (e) => {
+    e.preventDefault();
+    (window as any).PopUp.sendMessage({ type: 'stop', service: 'dropdown', respond: true });
```
Phenome: Maybe instead of `window as any`, add to `index.d.ts`:
```js
interface Window {
  PopUp: any;
}
```
(for now. Then we improve its definition later)

Phenome:
There's a "FOUC-like" flash that happens when you press play on a previous TE. More visible on FFox. Would be nice to tackle in #1291, together with some other stuff mentioned :+1:

PR #1366, opened 2019-03-29
feat: Add favro integration

timmygee:

Adds the timer integration to cards in the project management tool favro.com

### DIFF
--- a/src/scripts/origins.js
+++ b/src/scripts/origins.js
@@ -637,5 +637,13 @@ export default {
     url: '*://app.vivifyscrum.com/*',
     name: 'VivifyScrum',
     clone: 'true'
+  },
+  'favro.com': {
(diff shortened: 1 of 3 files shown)


src/scripts/origins.js
```diff
@@ -637,5 +637,13 @@ export default {
     url: '*://app.vivifyscrum.com/*',
     name: 'VivifyScrum',
     clone: 'true'
+  },
+  'favro.com': {
```
tcrammond: Hmm, it probably isn't necessary to have two entries here.

Just the first entry with `*://*.favro.com/*` as the `url` should work fine.

nfultz: Does that require the `.` before `favro` though? Don't want to accidentally match `myfavro.com` or other domains that end in the same string.

PR #1379, opened 2019-04-11
fix: Wrike - getting correct title

samuelszabo:

Timer description is extracted from Title of page. On Inbox page (https://www.wrike.com/workspace.htm?acc=XXXXX#path=inbox), the title is "Inbox" which creates not a correct description.

I want to parse the title from `ws-task-title` tag, which seems to be always placed just once on a page and is correct.

Because of "future-proof", this tag is used only if:

- is just one on a page (in case, there will be more tasks titles)
- is not empty

As a fallback, the current solution (title tag) is used.


samuelszabo:
@shantanuraj Thanks. Done.

shantanuraj:
Incorporated in #1407

Issue #939, opened 2017-12-07
Newly created tags don't show in Toggl Button


ApocalypticFish:
Please follow this guide when creating a new issue. This speeds up the process of replicating the issue and finding a solution.

### Environment details Google Chrome 62.0.3202.94

* Operation system: Windows 10
* Extension version: Version (1.8.0)

### Steps to reproduce the behavior 
-Start a time entry with Toggl Button, or with Toggl Button integration
-Create a new tag and click the Add Button and done
-New tag is created in the web app but isn't searchable in Toggl Button or the Toggl Button Extensions

### Additional details

* screenshots or screencapture

ApocalypticFish:
case/598494

PR #1386, opened 2019-04-24
Best effort new edit forms / timer popup

tcrammond:

Updates the appearance/layout of:
* "Edit TE" screen (when clicking on running TE in the toolbar pop up)
* "Timer popup" pop up when you start a timer in an integration

Adds new features:
* When changing the duration in "Edit TE" screen, you may now use `human inputs` in the same way as webapp. **Probably some minor glitches here, please test.**

Known things that won't be fixed:
* Would love to change the tags popdown to be a list (like webapp) and show selected tags at the start, but we're out of time. Will make a feature out of it in future.
* I have noticed that text rendering seems to get butchered slightly in the Chrome browser popup *(compare: open settings, change url to `popup.html`)* - no idea how to fix it.

Screen Shot 2019-04-26 at 11 05 02

Screen Shot 2019-04-26 at 11 06 44

Screen Shot 2019-04-26 at 11 06 52

Screen Shot 2019-04-26 at 11 07 10

Functionality should be.. functional. Check that nothing has gone missing.

Popdowns function sensibly *(they do still miss some keyboard features from webapp etc.)*, duration input functions.

#1291

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,9 @@
     "@emotion/styled": "^10.0.10",
     "date-fns": "^1.30.1",
     "keycode": "^2.2.0",
+    "lodash.map": "^4.6.0",
+    "lodash.reduce": "^4.6.0",
+    "moment": "^2.24.0",
     "prop-types": "^15.7.2",
     "react": "^16.8.4",
     "react-dom": "^16.8.4"
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -56,32 +56,57 @@ window.TogglButton = {
     '<form autocomplete="off">' +
     '<a class="toggl-button toggl-button-edit-form-button {service} active" href="javascript:void(0)">Stop timer</a>' +
     '<a id="toggl-button-hide">&times;</a>' +
-    '<div class="toggl-button-row" id="toggl-button-duration-row">' +
-    '<input name="toggl-button-duration" tabindex="100" type="text" pattern="[\\d]{2}:[\\d]{2}:[\\d]{2}" title="Please write the duration in the \'hh:mm:ss\' format." id="toggl-button-duration" class="toggl-button-input" value="" placeholder="00:00" autocomplete="off">' +
-    '</div>' +
-    '<div class="toggl-button-row">' +
-    '<input name="toggl-button-description" tabindex="101" type="text" id="toggl-button-description" class="toggl-button-input" value="" placeholder="(no description)" autocomplete="off">' +
-    '</div>' +
-    '<div class="toggl-button-row">' +
-    '<input name="toggl-button-project-filter" tabindex="102" type="text" id="toggl-button-project-filter" class="toggl-button-input" value="" placeholder="Filter Projects" autocomplete="off">' +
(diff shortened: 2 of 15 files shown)


src/scripts/background.js
```diff
@@ -1986,6 +2011,9 @@ window.TogglButton = {
       console.log('Load content script: [' + file + ']');
     }
     browser.tabs.insertCSS(tabId, { file: 'styles/style.css' })
+      .then(function () {
+        return browser.tabs.insertCSS(tabId, { file: 'styles/edit-form.css' });
+      })
       .then(function () {
         browser.tabs.insertCSS(tabId, { file: 'styles/autocomplete.css' });
```
shantanuraj: For consistency and correcting the promise chain I'd recommend retuning here as well
```suggestion
        return browser.tabs.insertCSS(tabId, { file: 'styles/autocomplete.css' });
```

package.json
```diff
@@ -93,6 +93,9 @@
     "@emotion/styled": "^10.0.10",
     "date-fns": "^1.30.1",
     "keycode": "^2.2.0",
+    "lodash.map": "^4.6.0",
```
shantanuraj: Based on the usage we can probably use the native `map/reduce` instead
to save some bits on the package size

tcrammond:
> not sure on the human input part or maybe its a problem in how I'm reviewing it

It's having problems when anything other than `blur` happens. Sorting it 👌

TheSoloJake:
@bryanph Updated your above comment to reflect the correct Jake :)

SuperGeekJake:
We used `react-maskedinput`. See toggl/webapp#9980

tcrammond:
Thx, working on the tweaks. :+1:

About some of the bonus remarks:

> Might be nice to add a back button to opt out of editing the TE

Not adding more scope to this PR but will definitely make an issue for this feature to be added soon.

> Make sure only numbers can be inserted in the time field

It allows flexible "human duration inputs" now, as in webapp, so this isn't necessary. 

> In the time input field, make sure that backspace just resets the number under the cursor to "0"  .....

Would be lovely, but far too out of scope for the goals here *(and it's definitely not a quick pull-in-the-library fix; this form is far from using React 😭)*. Let's put it on the "future UX redesign" list.

tcrammond:
Thanks @bryanph, if you just find a moment to verify the scroll bars issue in Linux has been resolved, that would be great.

src/scripts/lib/autocomplete.js
```diff
@@ -458,6 +459,8 @@ ProjectAutoComplete.prototype.setProjectBullet = function (pid, tid, el) {
 
   let task;
 
+  console.log(pid, tid, el, elem);
```
shantanuraj: Unguarded log

tcrammond:
> After creating tags from within the tag popdown, saving, editing the TE again and opening the tag popdown again, the created tags are not in the list

Exists here https://github.com/toggl/toggl-button/issues/939 *(2017 😬 )*, thanks for raising it anew. Will schedule it to be worked on.

PR #1410, opened 2019-05-15
Fix duration input interval behaviour

tcrammond:

Tweaks behaviour of the "ticking" duration input you get when editing the running TE.

The interval should no longer live on forever in the background, stacking up every time you use the form.

I checked on the pomodoro interval, and that one seemed sane.

If you want to see the old behaviour, add some logging to the interval in `showCurrentDuration` on master.

### DIFF
--- a/src/scripts/popup.js
+++ b/src/scripts/popup.js
@@ -69,8 +69,6 @@ window.PopUp = {
               JSON.stringify(TogglButton.$latestStoppedEntry)
             );
           }
-        } else {
-          PopUp.showCurrentDuration(true);
         }
         if (!PopUp.$header.getAttribute('data-view')) {
           PopUp.switchView(PopUp.$menuView);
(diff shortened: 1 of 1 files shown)


src/scripts/popup.js
```diff
@@ -258,11 +256,13 @@ window.PopUp = {
...
 
     PopUp.durationChanged = false;
-    PopUp.showCurrentDuration(true);
+    PopUp.updateDurationInput(true);
   },
 
-  showCurrentDuration: function (startTimer) {
+  updateDurationInput: function (startTimer) {
     if (TogglButton.$curEntry === null) {
+      console.log('exiting early');
```
shantanuraj: unguarded log

src/scripts/popup.js
```diff
@@ -275,14 +275,18 @@ window.PopUp = {
     }
 
     if (startTimer) {
-      if (!PopUp.$timer) {
-        PopUp.$timer = setInterval(function () {
-          PopUp.showCurrentDuration();
-        }, 1000);
-      }
+      clearInterval(PopUp.$timer);
```
shantanuraj: Semantically I'd prefer a call to the stop method
```suggestion
     PopUp.stopDurationInput();
```

Issue #1444, opened 2019-06-26
Idle Detection Issue 


VeljkoTN:
* **OS version**: Windows 10
* **Browser version**: 74.0.3729.169
* **Extension version**: TogglButton/1.37.4

### 🐛 Describe the bug

A user reached out saying that the Idle Detection notification is not showing. I tested it out and from what I understand the extension should see if you have been idle for 5 or more minutes, I let a timer run with Idle Detection previously enabled and went AFK (without a mouse attached) and the notification that I was idle never showed.

### Expected behavior

The Toggl Button extension should show a notification that a user has been idle if it's been more than 5 minutes of no activity

### Steps to reproduce

1. Make sure Idle Detection is enabled
2. Start a timer
3. Do not touch your computer (maybe best to disconnect your mouse for better test accuracy)
4. Wait for more than 5 minutes and see if the notification shows

### Intecom

[Case link. 
](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2097020/conversations/22585164741)

VeljkoTN:
In addition, the user said they also use Hubstaff and that Hubstaff actually did detect that she was away for 10 minutes but the Toggl Button did not.

VeljkoTN:
Ok this appears to be a notification thing so I'm closing this issue. I got the notification but my Windows notifications were set to "priority only" and it was just hidden

TheSoloJake:
@VeljkoTN User reporting that "priority only" wasn't set on their end, so that's not the issue. 

To also add to the information the user has provided, they have also checked with their IT Department on any potential blocks, and there are none.

chukitow:
@TheSoloJake @VeljkoTN from what I see in the code, the idle notification won't be triggered until the user's state change from **idle** to **active**.  is this the expected behavior? 

If this is not the expected behavior and the notification should be shown regardless if the user's state is **active** or **idle**, I created a pull request to solve the issue https://github.com/toggl/toggl-button/pull/1452

megamaiku:
I am now experiencing this issue on macOS 12.4

mknet:
I'm facing a similar issue with Toggle Track Desktop 8.5.0 on macOS: The idle time starts 20 minutes after I left the desk. Is this a setting issue or a bug?

PR #1452, opened 2019-07-09
Trigger notification after 5 minutes

chukitow:
According to https://github.com/toggl/toggl-button/issues/1444, The idle notification should be triggered after 5 minutes, but it will only be triggered when the user's state change from **idle** to **active**. 

This PR attempts to Trigger the idle notification after 5 minutes of idle time regardless of the use state.
Also if the user's state change from **idle** to **active**  and the inactive period were more than 5 minutes, I believe this is what the desktop app does.

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -1417,13 +1417,19 @@ window.TogglButton = {
       };
     };
 
+    const FIVE_MINUTES = 5 * 60;
+
+    const shouldTriggerNotification = (seconds) => {
+      const offsets = [-2, -1, 0, 1, 2];
+      return offsets.some(offset => seconds % (FIVE_MINUTES + offset) === 0);
+    };
+
     if (TogglButton.$user && ['active', 'idle'].includes(state) && TogglButton.$curEntry) {
       // trigger discard time notification once the user has been idle for
       // at least 5min
-      const FIVE_MINUTES = 5 * 60;
       if (
         TogglButton.$lastWork.id === TogglButton.$curEntry.id &&
-        inactiveSeconds % FIVE_MINUTES === 0 ||
+        shouldTriggerNotification(inactiveSeconds) ||
         (state === 'active' && inactiveSeconds >= FIVE_MINUTES)
       ) {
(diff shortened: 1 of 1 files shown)


src/scripts/background.js
```diff
@@ -1410,20 +1410,29 @@ window.TogglButton = {
...
 
-    if (TogglButton.$user && state === 'active' && TogglButton.$curEntry) {
+    if (TogglButton.$user && ['active', 'idle'].includes(state) && TogglButton.$curEntry) {
       // trigger discard time notification once the user has been idle for
       // at least 5min
+      const FIVE_MINUTES = 5 * 60;
       if (
         TogglButton.$lastWork.id === TogglButton.$curEntry.id &&
-        inactiveSeconds >= 5 * 60
+        inactiveSeconds % FIVE_MINUTES === 0 ||
```
shantanuraj: This would only trigger exactly at 5 minutes or multiples of it, I don't that's the intended behaviour

chukitow: what would be the expected behaviour? @shantanuraj

shantanuraj: Checking for modulo assumes the callback will be executed precisely at the 5:00 minute mark
The callback is scheduled to exectue every 2 seconds, I think you should accord for some scheduling delays and accept a delta of at least +/- 2 seconds

chukitow:
@shantanuraj I just updated the PR considering the offsets that you suggested :)

shantanuraj:
@chukitow I merged your changes in #1466  https://github.com/toggl/toggl-button/pull/1466/commits/2f8298ae6236e2b41f7645d3d18fa80b26bb6e97

- Did some minor cleanup to conform with our commit conventions
- Removed the delta thingie I suggested, without maintaining local state multiple notifications were being triggered at short intervals doesn't offer any significant improvements on the cost of more code.
- Refactored the code a bit.
- Squashed the commits

Closing this.

chukitow:
Awesome! @shantanuraj let me know if there is anything else I can help with :D

shantanuraj:
That'll be it for now, we'll push a release to the stores in a few hours
Thanks for your contribution 🙇

PR #1460, opened 2019-07-24
fix(visualstudio): use correct container to render button

### LINKED ISSUE #1459: Duplicate buttons in Azure DevOps / Visual Studio.
visual studio / azure devops
visualstudio.com

### 🐛 Describe the bug

When using queries view ({company}.visualstudio.com/{project}/_queries/query/{query_id}/) the button sometimes gets rendered multiple times and sometimes doesn't render at all.

image

### Expected behaviour

Button should only be rendered once.

### Other details or context

It happens because there can be multiple containers since previously visited work items are left hidden in the DOM. So it looks like an easy fix, I'll work on a PR myself.

image

tplk:

Container DOM nodes in query view are cached so it's needed to get an active (visible) container to prevent rendering all buttons in the first container.

Should be tested with different task views (I only tested normal and queries views)

closes #1459

### DIFF
--- a/src/scripts/content/visualstudio.js
+++ b/src/scripts/content/visualstudio.js
@@ -23,13 +23,15 @@ function projectSelector () {
   return projectElement ? projectElement.textContent : '';
 }
 
-function descriptionSelector () {
-  const formIdElem = $('.work-item-form-id span');
-  const formTitleElem = $('.work-item-form-title input');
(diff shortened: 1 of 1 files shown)


tplk:
@shantanuraj hi, any idea when anyone will have time to check my PR?

tplk:
@LulzAugusto thanks!

PR #1482, opened 2019-08-20
fix(slack): Fix slack integration after UI updates

### LINKED ISSUE #1477: Slack integration broken
* **OS version**: W10
* **Browser version**: Chrome 67
* **Extension version**: 1.40.1

Slack

### 🐛 Describe the bug

Toggl Button doesn't appear in Slack

### Expected behaviour

Toggl Button should appear next to channel actions and message actions

### Steps to reproduce

1. Go to Slack
2. Look for the button 🔎

### Other details or context

[This class](https://github.com/toggl/toggl-button/blob/master/src/scripts/content/slack.js#L5) doesn't seem to exist any more, no idea about the message actions though 🤷‍♀ 

[User report](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/3155115/conversations/23228965140)

LulzAugusto:
Fix both places where the button should show up:
1. at the top next to the channel name
2. next to the buttons that show up when you hover messages

1. Load extension
2. Enable Slack in the permissions settings
3. Go to https://toggl.slack.com or any other Slack account
4. Make sure the button shows up at the top next to the channel name
5. Make sure it is showing up when you hover messages 

👍 

closes #1477

### DIFF
--- a/src/scripts/content/slack.js
+++ b/src/scripts/content/slack.js
@@ -1,15 +1,17 @@
 'use strict';
 /* global togglbutton, $ */
 
+const getTextContent = (element) => element ? element.textContent.trim() : '';
+
 togglbutton.render('[data-qa="channel_name"]:not(.toggl)', { observe: true }, () => {
   const placeholder = $('[data-qa="channel_header__buttons"]').firstChild.parentNode;
-  const projectName = $('.p-classic_nav__team_header__team__name').textContent.trim();
-  const description = $('[data-qa="channel_name"]').textContent.trim();
+  const projectName = $('.p-classic_nav__team_header__team__name');
+  const description = $('[data-qa="channel_name"]');
 
   const link = togglbutton.createTimerLink({
     className: 'slack',
-    description,
-    projectName,
+    description: getTextContent(description),
+    projectName: getTextContent(projectName),
     buttonType: 'minimal'
   });
 
@@ -31,7 +33,7 @@ togglbutton.render('.c-message--hover:not(.toggl)', { observe: true }, elem => {
   const buttonContainer = document.createElement('span');
   const link = togglbutton.createTimerLink({
     className: 'slack-message',
-    projectName: projectName.textContent.trim(),
+    projectName: getTextContent(projectName),
     description: description.textContent.trim(),
     buttonType: 'minimal'
   });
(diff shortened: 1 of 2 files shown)


src/scripts/content/slack.js
```diff
@@ -1,12 +1,10 @@
...
 
-togglbutton.render('#channel_name:not(.toggl)', { observe: true }, () => {
-  const placeholder = $('.channel_title_info');
-  const projectName = $('#team_name').textContent.trim();
-  const description = $('#channel_name')
-    .textContent.trim()
-    .replace(/^#/, '');
+togglbutton.render('[data-qa="channel_name"]:not(.toggl)', { observe: true }, () => {
+  const placeholder = $('[data-qa="channel_header__buttons"]').firstChild.parentNode;
+  const projectName = $('.p-classic_nav__team_header__team__name').textContent.trim();
```
almirbi: same as above, maybe make a curry function for these two lines (`?.textContent.trim()`) 🤷‍♂

src/scripts/content/slack.js
```diff
@@ -1,12 +1,10 @@
...
 /* global togglbutton, $ */
 
-togglbutton.render('#channel_name:not(.toggl)', { observe: true }, () => {
-  const placeholder = $('.channel_title_info');
-  const projectName = $('#team_name').textContent.trim();
-  const description = $('#channel_name')
-    .textContent.trim()
-    .replace(/^#/, '');
+togglbutton.render('[data-qa="channel_name"]:not(.toggl)', { observe: true }, () => {
+  const placeholder = $('[data-qa="channel_header__buttons"]').firstChild.parentNode;
```
almirbi: 🚨 this will fail the script if element not found, can we use optional chains around here?

src/scripts/content/slack.js
```diff
@@ -1,12 +1,10 @@
 'use strict';
 /* global togglbutton, $ */
 
-togglbutton.render('#channel_name:not(.toggl)', { observe: true }, () => {
-  const placeholder = $('.channel_title_info');
-  const projectName = $('#team_name').textContent.trim();
-  const description = $('#channel_name')
-    .textContent.trim()
-    .replace(/^#/, '');
+togglbutton.render('[data-qa="channel_name"]:not(.toggl)', { observe: true }, () => {
```
almirbi: This is the same as someone would create a script to rely on `data-test-id` on toggl DOM elements 😁

almirbi:
Also, if you click the button while the button is stopped (grey), and the popup opens, closing the popup without clicking add still starts the TE (or at least it's kept highlighted, not grey anymore). 

The UX rule is that if you cannot know what the button exactly does (e.g. in case of icon buttons with no hover or anything), then you shouldn't do an action that changes data before the user confirms somehow. Not sure if this is something that we should fix extension wide or is it just happening in this PR.

src/scripts/content/slack.js
LulzAugusto: Well, I figured these are less likely to change than element's classes. Of course, if the interface changes regardless of the IDs things will still break, but I guess this makes us a bit less prone to errors due to small changes.

PR #1486, opened 2019-08-23
chore(chrome): Increase minimum required version for Chrome

### LINKED ISSUE #1391: Increase minimum required Chrome version
### Describe the new feature

There are some rather old versions of supported browsers being used in some cases that we can't effectively support. We should bump up the minimum Chrome version to *at least* 63. Check the stats.

Firefox was already handled recently, but we didn't check Chrome.

LulzAugusto:

Adds `minimum_chrome_version` to the chrome manifest file. 

It seems this is all you need to do, even though I couldn't confirm. I tried previewing a draft release, but the webstore doesn't display the minimum required version in their UI. 

Does anyone with experience maybe know if this is it? 🤷‍♀ 

closes #1391

### DIFF
--- a/src/chrome-manifest.json
+++ b/src/chrome-manifest.json
@@ -1,6 +1,7 @@
 {
   "name": "Toggl Button: Productivity & Time Tracker",
   "short_name": "Toggl Button",
+  "minimum_chrome_version": "63",
   "manifest_version": 2,
   "description":
     "Puts a timer into any web tool and allows quick real time productivity tracking with all the data stored on your Toggl account",


shantanuraj:
https://developer.chrome.com/extensions/manifest/minimum_chrome_version
This says yes

LulzAugusto:
Yup, I ended up there as well.

shantanuraj:
Just verifying the build on browserstack, I think we might have to change "63" to a semver string

LulzAugusto:
This says we can use "63" https://developer.chrome.com/extensions/manifest/version 🤔

shantanuraj:
Verified against Chromium 62
image

and Chromium 63
image

LulzAugusto:
Awesome. Thanks for the fast review 🤗

PR #1505, opened 2019-09-12
fix(popup): Fix main ui update delay

### LINKED ISSUE #1503: Duration in main view updates with a delay
* **OS version**: MacOs 10.14.5
* **Browser version**: Chrome 76.0.3809.132 
* **Extension version**: 1.40.8

### 🐛 Describe the bug

When changing the duration and pressing "Done" the main view still shows old time for a few seconds and then updates

### Expected behaviour

The duration in main page should already be updated when user return to main view. If the sync fails or any other issues appear we will return the old value. Most of the times it won't fail and the new duration is nicely shown right away. This removes the feeling of slowness in this part of the app

### Steps to reproduce

1. Start timer in popup
2. Click on timer to edit the entry
3. Edit duration
4. Press "Done" to accept changes
5. See that it takes a second for the duration to update on the main view

### Other details or context

tb-duration-update

IndrekV:

This PR makes the main popup UI update right after user presses "Done". That way the UI is already up to date when returning to the main view.

The main issue with duration was that it had a timer and when returning to main view the timer would still work and wait 1 second to update. I added another render call just before starting the interval timer. 

- Start timer
- Edit entry description and duration
- Press "Done"
- Check if the data is updated right away without delay

closes #1503

### DIFF
--- a/src/scripts/components/Timer.tsx
+++ b/src/scripts/components/Timer.tsx
@@ -77,6 +77,7 @@ function TimerDuration ({ start }: { start: string }) {
   const [ duration, setDuration ] = React.useState(formatDuration(start));
 
   React.useEffect(() => {
+    setDuration(formatDuration(start));
     const timeoutId = setTimeout(() => {
       setDuration(formatDuration(start));
     }, 1000);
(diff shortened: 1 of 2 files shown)


IndrekV:
After deep research of more than 4 years and considering various ways we could move forward with this, our research group has decided to not move forward and close this amazing PR.

Issue #1422, opened 2019-05-20
Tag dropdown and then Project dropdown will not stay open in Extensions


ApocalypticFish:
* **OS version**: 10.14.4
* **Browser version**: Firefox 66.0.5
* **Extension version**: 1.35.6

#### Relevant integration (if any): Teamweek and Bitrise
https://app.teamweek.com/

### 🐛 Describe the bug
Press the Toggl integration button within an app
Click the Tag drop down
(It should immediately close itself again)
After doing this several times it began happening to the Projects drop down as well.
### Expected behaviour
To be able to select a tag or project
### Steps to reproduce
Listed above

### Other details or context
Conversation 21957864690
I reproduced this as well as the user and there's a GIF in the conversation if you're unable to reproduce.

lise-toggl:
[case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/3155115/conversations/22988136255)

gif showing how you can reproduce the issue:
![ezgif com-gif-maker](https://user-images.githubusercontent.com/50156618/61944860-03a31580-afe2-11e9-9c5e-e1b52949395b.gif)

DustinWoods:
This has been happening to for a while too. I'm using Firefox 69.0. It happens on Github and Asana pages, as well as the extension pop up. Happens more often than not.

DustinWoods:
![Peek 2019-09-18 10-52](https://user-images.githubusercontent.com/5541776/65172895-a5ebe200-da02-11e9-89a4-17e4b63f3cd4.gif)

**Workaround**: It seems if you right-click the tag drop down (or the project list dropdown) that the menu will stay open on the subsequent left click. See gif.

dooart:
there's a big chance this will be fixed by https://github.com/toggl/toggl-button/pull/1504

dooart:
nevermind, just tested it and the problem still happens. will schedule for next backlog cycle

DustinWoods:
@dooart Thank you! This one is a big deal for me.

DustinWoods:
I'd be happy to help troubleshoot this issue or gather more info if needed.

VeljkoTN:
conversations/23731381826 - video reproducing the issue included in one of the messages

DustinWoods:
Thank you!!! I'll report back here after I update the extension in my browser.

dooart:
[...]

PR #1517, opened 2019-10-04
fix(popup): Fix tag/project dropdown toggle in-page

nunofmn:
- 0c0a6e5 - **fix(popup): Fix tag/project dropdown toggle in-page**

Similar fix to d40b843, but this time for the dropdown in-page

 - Check if tab/project popdowns in the extension works fine. Inside integrations (e.g. Trello) as well as from the main toolbar popup.

Related #1422

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -92,6 +92,7 @@ window.TogglButton = {
     <div class="TB__Dialog__field" tabindex="0">
       <div>
         <div id="toggl-button-project-placeholder" class="TB__FormFieldTrigger__trigger" disabled><span class="tb-project-bullet"><div>No project</div></span><span class="TB__Popdown__caret"></span></div>
+        <div class="TB__Popdown__overlay"></div>
         <div class="TB__Popdown__content">
           <div class="TB__Popdown__filterContainer">
             <input name="toggl-button-project-filter" type="text" id="toggl-button-project-filter" class="TB__Popdown__filter" value="" placeholder="Find project..." autocomplete="off">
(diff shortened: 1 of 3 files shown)


src/scripts/background.js
```diff
@@ -85,7 +85,6 @@ window.TogglButton = {
     <div class="TB__Dialog__field" tabindex="0">
       <div>
         <div id="toggl-button-project-placeholder" class="TB__FormFieldTrigger__trigger" disabled><span class="tb-project-bullet"><div>No project</div></span><span class="TB__Popdown__caret"></span></div>
-        <div class="TB__Popdown__overlay"></div>
```
dpsxp: Removed the overlay because it wasn't really necessary and was breaking the capability of clicking on other elements on the page.

src/scripts/lib/autocomplete.js
```diff
@@ -31,12 +32,17 @@ const AutoComplete = function (el, item, elem) {
...
 
+  let dropdownRaceTimer;
   that.placeholderItem.addEventListener('click', function (e) {
-    setTimeout(function () {
+    clearTimeout(dropdownRaceTimer);
+
+    dropdownRaceTimer = setTimeout(function () {
       that.toggleDropdown();
-    }, 50);
+    }, 150);
```
tcrammond: Hmm this delay feels quite unpleasant. Do you think it can be slightly lower or maybe there's another way?

dpsxp:
Ok, So I took another good look at this to see if we could remove the timer. I wasn't happy with the solution too, the timer can be different on each machine so the problem could happen again 😞 .

Now there's no time and the race condition should be fixed. Instead of relying on timings I'm checking if the element that is causing the duplicated `focus` event is inside the pop-down if so there's no need to change the pop-down state.

tcrammond:
Update: still having problems in fresh Chrome profile. It's fine on Firefox though. Apparently it's totally fine in Chrome for @shantanuraj 🤔

tcrammond:
My gut feeling suggestion is to bring the overlay back @dpsxp and go from there 🤔

dpsxp:
Ok, I was able to reproduce the issue. Did some changes, it should be ready for another review now.

PR #1518, opened 2019-10-06
feat: add zammad integration for toggl button

athanasius-kircher:

I request the merge of the zammad integration. Zammad is a helpdesk and support software see  https://zammad.com.

The toggl button is placed in the ticket details in the tool-area at the top right (see screenshot below).

The changes have been linted and the integration has been checked in Chrome and Firefox.

zammad

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Add Toggl one-click time tracking to popular web tools.
 - [Wunderlist](https://www.wunderlist.com)
 - [Xero](https://www.xero.com/)
 - [YouTrack](http://www.jetbrains.com/youtrack/)
+- [Zammad](https://zammad.com/)
 - [Zendesk](https://www.zendesk.com/)
 - [Zube](https://zube.io/)
 - [VivifyScrum](https://www.vivifyscrum.com/)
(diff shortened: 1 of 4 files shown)


lise-toggl:
Hey @athanasius-kircher, thank you for the PR!

The button on the looks good.

However, for me, the project selector always returns "Zammad Helpdesk". Is this correct? Would it be possible to change it to the customer name instead?

I also think it would be good to have the mini button appear on the overview list:
image

athanasius-kircher:
Ok I have created a new commit. Changes to the version before and the questions above:
1. The project string is not taken from the site name anymore, but is a combination of the client name and the organization. The later is only taken if the client is associated to an organization. The project string can be described by the following pseudo code `client.fullName + ((organization.exists)?' (' + organization.name + ')')`. This pattern resemble the Zammad client name pattern in different lists.
2. @lise-toggl I have checked the options to integrate the button into the overview list, but can not see a good solution. The problem is, that the different lists are very unspecific and relevant information are missing. The lists use very generic markup (overlapping with lists for users etc.) and do not have very reliable selectors, as it seems. Some ticket list use a special wrapper `content_permanent_TicketOverview`. But on the other hand ticket-results in a search are quite difficult to distinguish from users for example. And even in the overview list the public ticket id is missing and does not occure in the markup. Therefore I would propose to do not implement the button in the list (currently), because it may lead to more tracking inconsistency due to different identifiers or different list behaviour.

lise-toggl:
Great, thank you for the update! :)

PR #1533, opened 2019-10-25
fix(Todoist): Fix time entry edit form on dark mode

dooart:

Follow up to #1532.

There were more elements not visible or just weird because of Todoist's dark mode. These fixes should improve things in other tools using dark mode as well.

Make sure Todoist works in both "Todoist" (default) and "Dark" themes. I also tested with Github but It'd be nice to check in a third tool just in case.

List of things fixed by this PR (all in Dark theme):

1. Label for the stop timer button inside the edit form didn't appear:
image

2. Add new tag button looked weird:
image

3. No placeholder in popdown filter inputs:
image

### DIFF
--- a/src/styles/autocomplete.css
+++ b/src/styles/autocomplete.css
@@ -5,6 +5,10 @@
   padding: 0 6px;
 }
 
+#tag-autocomplete::-webkit-scrollbar-track, .TB__Popdown__content::-webkit-scrollbar-track {
+  background: initial !important;
+}
+
 #tag-autocomplete {
   /* aligned with filter bar whilst it's the "old" tag design as opposed to a list */
   padding: 0 15px;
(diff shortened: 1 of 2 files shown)


shantanuraj:
Okay just for reference, the height bug isn't present in 1.44.1 currently rolled out

dooart:
Yeah that's because Todoist specifically adds a CSS rule to make it thicker, I just decided to leave it like that.

Btw I was thinking it might make sense to embrace the dark mode on our side as well. What I don't know is how to do that using only CSS selectors in a way that would also work with other tools.

shantanuraj:
There’s some work in the settings page that you can look at.

tcrammond:
Is this good to go?

dooart:
ugh i thought this was in already

PR #1552, opened 2019-11-07
fix(any.do): Update description and project selectors

### LINKED ISSUE #1551: Any.do: Description not fetched in list view
* **OS version**: Ubuntu 19.10
* **Browser version**: Chrome 78
* **Extension version**: 1.45

[Any.do](https://desktop.any.do/tasks/)

### 🐛 Describe the bug

In the tasks list, the description is not fetched when starting a timer:
anydo

### Expected behaviour

The description should be fetched

### Steps to reproduce

1. Go to Any.do
2. Add a task
3. Start a timer for that task in the list view

### Other details or context

Found while working on #1550

### LINKED ISSUE #1553: Improve Any.do project selection
### Describe the changes you'd like

Select the correct project when possible in Any.do

### Relevant links or screenshots

When moving from the "All Tasks" page to another project's page, the task list items don't update their project (i.e. still has "All Tasks" set as their project).

lise-toggl:

Gets the correct task description from Any.do's task list.

Track an item from the task list and check the correct description is used.

Closes #1551 

Also closes #1553

### DIFF
--- a/src/scripts/content/anydo.js
+++ b/src/scripts/content/anydo.js
@@ -3,24 +3,18 @@
 
 // any.do Q4 2018: task card
 togglbutton.render('.CardScrollView:not(.toggl)', { observe: true }, elem => {
-  const descriptionElem = $('textarea', elem);
+  const descriptionSelector = () => elem.querySelector('textarea').value;
 
-  // Since task popup has same selector as overview
-  // do nothing when description is not present
-  if (!descriptionElem) {
-    return;
-  }
-
-  const projectElem = $('.TasksToolBarCategoryTitle');
+  const projectSelector = () => document.querySelector('.TasksToolBarCategoryTitle').textContent;
 
   const link = togglbutton.createTimerLink({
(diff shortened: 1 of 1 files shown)


src/scripts/content/anydo.js
```diff
@@ -28,13 +28,13 @@ togglbutton.render(
   '.TaskListRow[draggable=true]:not(.toggl)',
   { observe: true },
   elem => {
-    const descriptionElem = $('input', elem);
+    const descriptionElem = elem.querySelector('.TaskItem__label__text__title').textContent;
```
tcrammond: We should change this into a function that returns the description string, like we do in more modern integrations (usually called something like `descriptionSelector` if you need a reference)

This will mean that we'll always grab the latest description, even after the user might have changed it in the `any.do` UI.

lise-toggl: Thanks! I've updated all of them to work this way :+1:

Issue #1179, opened 2018-11-06
Active TE name not syncing with webapp


bryanph:
Please follow this guide when creating a new issue. This speeds up the process of replicating the issue and finding a solution.

### Environment details

* Operation system: Firefox
* Extension version: Latest (after #1174)

### Expected behavior
The description should be synced and changed in the webapp as well

### Actual behavior
It doesn't get synced
![te-bug1](https://user-images.githubusercontent.com/3816712/48056124-684ae580-e1b1-11e8-8c6a-b2e8cf4072f2.gif)

### Steps to reproduce the behavior
See the GIF

### Additional details

* screenshots or screencapture

langri-sha:
Could be it was something transient with sockets or the timer form. Worth taking another look.

PR #1556, opened 2019-11-12
Enable websocket reconnection attempts

tcrammond:

Syncing issues are cropping up between button and other clients. Websocket being closed and then not reconnecting is a likely cause.

Let's reimplement reconnection attempts.

Logic at time of opening PR:
- Each retry waits 5 seconds longer than the last.
- The maximum retry wait time is 1 minute
- The maximum number of retries is 30, at which point it gives up

Ideally we'd have some way to rekindle it in the UI, but it requires some further thought. I would like to merge this without adding on new features first.

I do think there really needs to be some way to easily kickstart it again after it goes  wrong.

To test the continual reconnection logic, I recommend:
- Set the reconnection interval to something short like `1`
- Set the maximum attempts to something small like `5`
- Change the websocket url to an invalid host
- Reload the extension. You will now see logging output in the background page inspector. After its retries it will stop.

At any time in the background page inspector, you can `TogglButton.websocket.close()` to close the connection and trigger a reconnect. All of the state is reset when reconnection is successful, so you'll normally only ever see one reconnect.

Related to #1179

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -296,7 +296,7 @@ window.TogglButton = {
     });
 
     TogglButton.websocket.socket.onopen = function () {
-      TogglButton.resetWebsocketState();
+      TogglButton.resetWebsocketRetryCount();
       const data = JSON.stringify(authenticationMessage);
       try {
         return TogglButton.websocket.socket.send(data);
@@ -358,7 +358,7 @@ window.TogglButton = {
   },
 
   // Resets the reconnection state to give things another chance
-  resetWebsocketState: function () {
+  resetWebsocketRetryCount: function () {
     TogglButton.websocket.retryCount = 0;
   },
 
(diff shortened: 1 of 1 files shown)


src/scripts/background.js
```diff
@@ -48,8 +48,15 @@ window.TogglButton = {
...
   $ApiV9Url: `${process.env.API_URL}/v9`,
   $sendResponse: null,
-  $socket: null,
-  $retrySocket: false,
+  websocket: {
+    socket: null,
+    maxRetryCount: 30,
+    retryCount: 0,
+    retryInterval: 5,
+    get retryEnabled () {
```
shantanuraj: Dishy 👏

src/scripts/background.js
```diff
@@ -286,28 +295,28 @@ window.TogglButton = {
       type: 'pong'
     });
 
-    TogglButton.$socket.onopen = function () {
+    TogglButton.websocket.socket.onopen = function () {
+      TogglButton.resetWebsocketState();
```
shantanuraj: My only suggestion would be some alternate name, this stood out to me since I thought it resets the actual websocket connection as soon the socket connection is opened

Maybe `resetWebsocketRetryCount`  or something

PR #1557, opened 2019-11-13
fix(popup): Add minimum height to popup

### LINKED ISSUE #1534: Toolbar window: Minimum height has gone missing
### 🐛 Describe the bug

The toolbar window now looks like this with no recent entries:

Screenshot 2019-10-28 at 11 04 17

### Expected behaviour

It should have some kind of height, it used to be `300px` or something like that, I think. A change to the way browser extensions work has probably broken it.

I believe this is usually set in the Manifest, but I can't quite remember. Maybe it was a minimum height in the content.

### Steps to reproduce

1. Install in Chrome
2. Log into account with no recent entries or few recent entries
3. See that the window is not very tall when you open it.

lise-toggl:

Specifies a minimum height for the Toggl Button popup so that it doesn't look all squashed for accounts with no or few entries:
image

1. Login to the Button with an account with 0 entries

Closes #1534.

WIP: waiting for clarification on whether the "See more on ..." footer should always be displayed or not (or some other placeholder message for accounts without entries).


shantanuraj:
I think we always want to display the see more, it opens the webapp and can be used to do stuff not possible with the extension, eg. reports

lise-toggl:
Done, now always displays the footer message:

image

shantanuraj:
Let's push the See more button to the bottom
and to make some use of the middle section  let's reuse webapp's empty state onboarding image and copy

## webapp
image

#### Image
toggl-stopwatch

#### Copy
"Get ready to track time and boost your productivity!"

src/scripts/components/TimeEntriesList.tsx
```diff
@@ -35,11 +35,23 @@ const getTimeEntryDayGroups = (timeEntries: Array<Array<Toggl.TimeEntry>>): {[da
...
+    </a>
+  </Footer>);
+
 interface TimeEntriesListProps {
   timeEntries: Toggl.TimeEntry[][];
   projects: IdMap<Toggl.Project>;
 };
-export default function TimeEntriesList (props: TimeEntriesListProps) {
+export default function TimeEntriesList (props?: TimeEntriesListProps) {
+  if(Object.keys(props).length === 0) {
```
shantanuraj: TS is throwing an error here

shantanuraj: PS. I think you don't really need this

src/scripts/components/TimeEntriesList.tsx
```diff
@@ -35,11 +35,23 @@ const getTimeEntryDayGroups = (timeEntries: Array<Array<Toggl.TimeEntry>>): {[da
...
+      </Button>
+    </a>
+  </Footer>);
+
 interface TimeEntriesListProps {
   timeEntries: Toggl.TimeEntry[][];
   projects: IdMap<Toggl.Project>;
 };
-export default function TimeEntriesList (props: TimeEntriesListProps) {
+export default function TimeEntriesList (props?: TimeEntriesListProps) {
```
shantanuraj: props will never be `undefined`, so if you want to mark things as optional mark the keys in L48 and L49

src/scripts/components/TimeEntriesList.tsx
shantanuraj: Line 55 below deals with this, it sets `timeEntries` to an empty array, so you move the if statement after it
and change it something like

```typescript
if (timeEntries.length === 0) {
  return <TimeEntriesFooter />;
}
```

shantanuraj: Though in this case you might want something like

```typescript
if (timeEntries.length === 0) {
  return (
    <React.Fragment>
      <EmptyState />
      <TimeEntriesFooter />
    </React.Fragment>
  );
}
```

lise-toggl:
How's this?

image

image

shantanuraj:
Perfect 👌

src/scripts/components/TimeEntriesList.tsx
lise-toggl: updated :raised_hands:

src/scripts/components/TimeEntriesList.tsx
lise-toggl: reverted :raised_hands:

lise-toggl:
Or maybe we could move down the message a bit? :eyes:

shantanuraj:
That works as well 👍

lise-toggl:
Done ＼(★^∀^★)／

image

PR #1560, opened 2019-11-14
Implement exponential backoff

shantanuraj:

- Max limit is an hour
- No limits on retries

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -7,6 +7,7 @@ import Ga from './lib/ga';
 const browser = require('webextension-polyfill');
 const FIVE_MINUTES = 5 * 60;
 const ONE_HOUR = 60 * 60;
+const RETRY_INTERVAL = 15;
 
 let openWindowsCount = 0;
 
@@ -24,13 +25,13 @@ function randomInt (min, max) {
 }
 
 function jitter () {
-  // Random amount between 0-5 seconds
-  return randomInt(0, 5);
+  // Random duration between 0-60 seconds
+  return randomInt(0, 60);
 }
 
-function backoff (retryInterval, retryCount) {
+function backoff (retryCount) {
   return Math.min(
-    (retryInterval * Math.pow(2, retryCount)) + jitter(),
+    (RETRY_INTERVAL * Math.pow(2, retryCount)) + jitter(),
     ONE_HOUR
   );
 }
@@ -67,9 +68,7 @@ window.TogglButton = {
   $sendResponse: null,
   websocket: {
     socket: null,
-    maxRetryCount: 30,
-    retryCount: 0,
-    retryInterval: 5
+    retryCount: 0
   },
   $nannyTimer: null,
   $lastSyncDate: null,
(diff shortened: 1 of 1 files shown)


src/scripts/background.js
```diff
@@ -52,11 +69,7 @@ window.TogglButton = {
     socket: null,
     maxRetryCount: 30,
     retryCount: 0,
-    retryInterval: 5,
-    get retryEnabled () {
-      return false;
-      // return TogglButton.websocket.retryCount < TogglButton.websocket.maxRetryCount;
-    }
+    retryInterval: 5
```
leafi: In the case of the connection repeatedly working, then closing, then working, then closing..., we'd be retrying every 5s after the connection closes. IMO this is too soon.

Suggest retryInterval should be 30 seconds

src/scripts/background.js
```diff
@@ -18,6 +19,22 @@ const shouldTriggerNotification = (state, seconds) => {
   );
 };
 
+function randomInt (min, max) {
+  return Math.floor(Math.random() * (max - min + 1)) + min;
+}
+
+function jitter () {
+  // Random amount between 0-5 seconds
+  return randomInt(0, 5);
```
leafi: I think, given the range of users we have on different types of connection, it's already true there's about 5 seconds of jitter (though most users would probably take 0-3 seconds to connect, not 5 seconds).

What do you think about changing this range from 0..5 to 0..30, or even 0..60? IMO 5s jitter is very little

src/scripts/background.js
shantanuraj: It is currently 5 + rand(0, 5) in the above case

src/scripts/background.js
```diff
@@ -18,6 +19,22 @@ const shouldTriggerNotification = (state, seconds) => {
...
+}
+
+function backoff (retryInterval, retryCount) {
+  return Math.min(
+    (retryInterval * Math.pow(2, retryCount)) + jitter(),
```
leafi: On 1st retry (so we've failed to connect once), retryCount is 0, correct? That's how I think the code works.. and in that case this is correct. Just want to confirm with you.

src/scripts/background.js
shantanuraj: I'd bump either the base amount or jitter, leaning towards jitter

src/scripts/background.js
```diff
@@ -341,21 +350,15 @@ window.TogglButton = {
...
-
-        bugsnagClient.leaveBreadcrumb(`Websocket reconnection attempt ${TogglButton.websocket.retryCount}`);
-        if (process.env.DEBUG) console.info(`Websocket reconnection attempt ${TogglButton.websocket.retryCount}`);
-
-        if (!TogglButton.websocket.retryEnabled) {
-          bugsnagClient.leaveBreadcrumb('Websocket reconnections abandoned');
-        }
-      }, retrySeconds * 1000);
-    }
+    // Retry connection, increasing the timeout each time, up to 30 minutes.
```
leafi: comment says 30 min but code above says ONE_HOUR

src/scripts/background.js
shantanuraj: Yep so 5 + 0 + rand(0, 5)

src/scripts/background.js
shantanuraj: Will fix 🙇

src/scripts/background.js
leafi: Would still rather look at e.g. 30s + rand(0, 60s), given that base 'retryInterval: 5' is current behavior & we are demonstrably causing problems for BE

shantanuraj: do you think

base retry: 15 seconds
max_jitter: 60 seconds

want to keep the regular user which might get disconnected on closing laptop lid happy

so yeah worst case time in first attempt

15 + 0 = 15 seconds after they open/resume their machines

in next attempt worst case is already 30 seconds now

15 * 2 + 0 = 30 seconds

in both cases assuming jitter is randomly 0

leafi:
Discussed numbers with shantanu in PR... awaiting new push, then will approve

leafi:
Code LGTM - trying to install this extension now

PR #1590, opened 2019-11-21
feat(focus): Add focus mode for pomodoro

### LINKED ISSUE #1567: [Phoenix] Add a clock view for running Pomodoro entry
- Make it optional
- Only render the current running entry in a circular countdown clock
- Add stop entry button

image

lise-toggl:

Adds an optional focus mode while a pomodoro timer is running - shows a countdown timer instead of the usual entries list.

Current issues:
- The countdown doesn't update at the exact same time as the duration :(
- It works in the extension popup, but not when the popup is opened as a separate page

Turn the focus mode option on and start a timer.
Click on the countdown to stop it early.
Open the popup during a working 

Closes #1567

### DIFF
--- a/src/html/settings.html
+++ b/src/html/settings.html
@@ -208,6 +208,10 @@
                     <input type="checkbox" id="pomodoro-stop-time">
                     <label for="pomodoro-stop-time">Stop time tracking when pomodoro time ends</label>
                   </p>
+                  <p>
+                    <input type="checkbox" id="pomodoro-focus-mode">
+                    <label for="pomodoro-focus-mode">Enable focus mode while a pomodoro interval is active</label>
+                  </p>
                   <div class="pomodoro-sound">
                     <input type="checkbox" id="enable-sound-signal">
                     <label for="enable-sound-signal">Enable sound notification at the end of an interval</label>
--- a/src/scripts/@toggl/time-format-utils/format-duration.ts
+++ b/src/scripts/@toggl/time-format-utils/format-duration.ts
@@ -1,0 +1,20 @@
+import {
+  differenceInHours,
+  differenceInMinutes,
+  differenceInSeconds,
+  subHours,
+  subMinutes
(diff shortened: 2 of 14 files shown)


shantanuraj:
Rebased changes on latest release branch

lise-toggl:
Re: workaround if you think it's ok to not display the duration that sounds like a plan :)

If we invert the colours (like this idea) I think the cursor should change on the whole circle, and the whole circle should be clickable. What do you think?

shantanuraj:
Didn't invert color, the big red-circle looked too jarring 

Fixed the problems stated in OP

Let's do a round of review to ensure

Focus mode only kicks in in pomdoro enabled and focus preference is turned on as well

shantanuraj:
Didn't invert color, the big red-circle looked too jarring 

Fixed the problems stated in OP

Let's do a round of review to ensure

Focus mode only kicks in in pomdoro enabled and focus preference is turned on as well

shantanuraj:
Didn't invert color, the big red-circle looked too jarring 

Fixed the problems stated in OP

Let's do a round of review to ensure

Focus mode only kicks in in pomdoro enabled and focus preference is turned on as well

lise-toggl:
I found the duration fades each time you open the popup, is this intentional?
cut

Also, did you want to limit the cursor style and stop on click to the red rectangle only?

shantanuraj:
@lise-toggl I've made some more improvements regarding the render
Extracted and refactored the timer code used in react components to hooks, scheduling next render using `requestAnimationFrame`. 
This is ready for review and hopefully merge now 🙇

shantanuraj:
You can ignore changes made to `timerUtils.js` in 34ed96845313ef021a4008b55acf4b7849256d48
Just applied some lint fixes VS Code was complaining of

lise-toggl:
very nice :clap:

PR #1539, opened 2019-10-29
Fixed #1538

nopantsmonkey:
Please remember the [Contributing Guidelines](https://github.com/toggl/toggl-button/blob/master/.github/CONTRIBUTING.md) :heart:

## :star2: What does this PR do?
Fixed issue #1538 

undefined checks updated.

## :bug: Recommendations for testing

All changes should be tested across Chrome and Firefox.

PR #1602, opened 2019-12-02
fix(popup): Stop time cleared when opening running entry

lise-toggl:
#1539  :star2: What does this PR do?

Fixes a bug where editing the running entry also stops it, if a completed entry has been opened previously. 

Steps to reproduce:
- Open edit view for any past entry
  - `dataset.stopTime` is set to that entry's stop time
- Open edit view for the currently running entry
  - `dataset.stopTime` is still set to the previous stop time
  - Submitting the form stops the entry

This change clears `dataset.stopTime` when opening the currently running entry.

- Follow the instructions above to verify the bug in master
- Try the same again in this branch

### DIFF
--- a/src/scripts/shared/use-duration.ts
+++ b/src/scripts/shared/use-duration.ts
@@ -12,7 +12,7 @@ export const useDuration = (start: string) => {
 
   const updateDuration = React.useCallback(() => {
     setDuration(timeElapsed(start));
-  }, [setDuration]);
+  }, [start, setDuration]);
 
   useInterval(updateDuration, ONE_SECOND);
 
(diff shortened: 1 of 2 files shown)


shantanuraj:
Your fix works well there's another issue though

When you edit the running entry all changes are reflected in the list-view
except if you edit the duration of the running entry, on saving the change
the new duration isn't shown in the running timer bar

BE has the right duration, so we are sending the right data, and if you stop the entry, the correct duration is shown, we'll have to look into the timer logic I guess

Dec-02-2019 11-10-12

lise-toggl:
The duration not updating happens to me on master as well, I'll have a look at this now :thinking:

lise-toggl:
Think I've fixed it!

src/scripts/shared/use-duration.ts
```diff
@@ -12,7 +12,7 @@ export const useDuration = (start: string) => {
 
   const updateDuration = React.useCallback(() => {
     setDuration(timeElapsed(start));
-  }, [setDuration]);
+  }, [duration]);
```
shantanuraj:
```suggestion
  }, [setDuration, start]);
```
Can you verify with this?

PR #1612, opened 2019-12-04
Fix Notion integration

### LINKED ISSUE #1454: Notion Integration Isn't appearing.
* **OS version**:  MacOS 10.14.5
* **Browser version**: 75.0.3770.100
* **Extension version**: 1.37.4

#### Relevant integration (if any):Notion

www.notion.so

### 🐛 Describe the bug
The Toggl button integration isn't appearing within Notion

Some other users have also mentioned it [here](https://github.com/toggl/toggl-button/issues/896) and [here](https://github.com/toggl/toggl-button/pull/1203) also in this [case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/conversation/22009910172)

### Expected behaviour

The Toggl button won't appear in Notion

### Steps to reproduce
A Togglr opens Notion and looks for the Toggl button.
They scream for it is not here and thus they cannot Toggl.

rylek90:

Fixes Notion integration for some (hopefully longer) time, until it changes DOM structure again 😅 

Toggl button should appear in dialog view as visible below:
image

Closes #1454

### DIFF
--- a/src/scripts/content/notion.js
+++ b/src/scripts/content/notion.js
@@ -1,27 +1,24 @@
 'use strict';
 
+// Selectors here are madness, it works for as of Dec 4th 2019
+// Button renders in popup/dialog view
 togglbutton.render(
-  '.notion-overlay-container .notion-peek-renderer .notion-page-controls + .notion-selectable > [contenteditable="true"][placeholder="Untitled"]:not(.toggl)',
+  '.notion-peek-renderer:not(.toggl)',
   { observe: true },
   function (elem) {
-    const container = createTag('div', 'button-link notion-tb-wrapper');
-    const descriptionElem = elem;
-    const projectElem = $(
-      '.notion-sidebar-container > * > * > * > * > * > * > * > * + * > * > *'
-    );
-    const togglButtonLoc = $(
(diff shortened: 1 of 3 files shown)


lise-toggl:
hey @rylek90 it's pretty hard to read on dark mode:
image

and, iirc the button used to be on the main page as well, not just popups, do you remember if that was the case?

rylek90:
Thanks @lise-toggl 🙌 
* Fixed dark mode appearance
* Added integration on main page

lise-toggl:
for some reason the :last-child selector didn't work for me, but using `elem.lastChild.lastChild` does select the correct action block on the main page.

there's also another issue though - the link doesn't get updated on page change :thinking:

rylek90:
> Sorry, the main page button goes in the wrong place (see left breadcrumbs) when loading a page (and not sure it works during regular navigation)

Seems like race condition, I was able to reproduce it after few tries

> If it's looking difficult by itself, we can handle it separately from this PR and not include it if you prefer.

We can try position button in hover menu (see below) in the future but let's merge it first in popup mode only.
image

> there's another issue though - the link doesn't get updated on page change :thinking: 

Yup, Notion doesn't refresh this element on page change. We need to regroup and think of better approach. For now let's leave it since it already quite better than before

PR #1617, opened 2019-12-11
Add button to Notion main page

lise-toggl:

adds timer link to notion main page (extends #1612)
- wasn't able to get the hover text to update when the page updates, but the timer will start with the correct description :(

also adds a wrapper to make the timer link look more like native page actions

try out the timer from the notion main page
check the new style looks ok

follow on from #1612 for #1454

### DIFF
--- a/src/scripts/content/notion.js
+++ b/src/scripts/content/notion.js
@@ -1,24 +1,23 @@
 'use strict';
 
-// Selectors here are madness, it works for as of Dec 4th 2019
-// Button renders in popup/dialog view
+// Button renders left of page title
 togglbutton.render(
-  '.notion-peek-renderer:not(.toggl)',
+  '.notion-page-controls + div:not(.toggl)',
   { observe: true },
   function (elem) {
+    elem.style.position = 'relative';
+
     function getDescription () {
-      const descriptionElem = elem.querySelector('.notion-scroller .notion-selectable div[contenteditable="true"]');
+      const descriptionElem = elem;
       return descriptionElem ? descriptionElem.textContent.trim() : '';
     }
 
     const link = togglbutton.createTimerLink({
       className: 'notion',
+      buttonType: 'minimal',
       description: getDescription
     });
 
-    const root = elem.querySelector('div:nth-child(2) > div:nth-child(1) > div:nth-child(1) > div:nth-child(3)');
-    if (root) {
-      root.prepend(link);
-    }
+    elem.prepend(link);
   }
 );
--- a/src/scripts/content/notion.js
+++ b/src/scripts/content/notion.js
@@ -1,6 +1,33 @@
 'use strict';
 
-// Button renders left of page title
+// Selectors here are madness, it works for as of Dec 4th 2019
+// Button renders in popup/dialog view
+togglbutton.render(
+  '.notion-peek-renderer:not(.toggl)',
+  { observe: true },
+  function (elem) {
+    function getDescription () {
+      const descriptionElem = elem.querySelector('.notion-scroller .notion-selectable div[contenteditable="true"]');
+      return descriptionElem ? descriptionElem.textContent.trim() : '';
+    }
+
+    const link = togglbutton.createTimerLink({
+      className: 'notion',
+      description: getDescription
+    });
+
+    const wrapper = document.createElement('div');
+    wrapper.classList.add('toggl-button-notion-wrapper');
+    wrapper.appendChild(link);
+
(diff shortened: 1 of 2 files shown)


src/scripts/content/notion.js
```diff
@@ -16,9 +16,40 @@ togglbutton.render(
...
+      className: 'notion',
+      description: getDescription
+    });
+
+    const wrapper = document.createElement('div');
+    wrapper.classList.add('toggl-button-notion-wrapper');
+    wrapper.appendChild(link);
+
+    const root = elem.querySelector(':scope > div > div:last-child');
+    console.log(root.querySelector(':scope > div').textContent);
```
rylek90: Please remove me 😄

src/scripts/content/notion.js
```diff
@@ -16,9 +16,40 @@ togglbutton.render(
...
+  }
+);
+
+// Button renders on main page
+togglbutton.render(
+  '.notion-topbar:not(.toggl)',
+  { observe: true },
+  function (elem) {
+    function getDescription () {
+      const descriptionElem = elem.querySelector(':scope > div > div > div:last-child');
```
rylek90: Do we need page icons/emojis in description? 🤔 

image

src/scripts/content/notion.js
lise-toggl: :see_no_evil:

src/scripts/content/notion.js
lise-toggl: thanks - I also noticed it gets cut off if the description is too long, so I've changed a better selector here

lise-toggl:
@rylek90 in the end I found it was easier to have the button display as part of the page, instead of as part of the header :(

Since this works for both the main page and the popup, I've replaced the popup specific attachment with this one.

On positioning - I wanted to make the button appear after the text like usual for the minimal button, but I ran into issues when the page title has multiple lines. I tried putting the button itself inside the title div, which makes it appear just after the title text, but then the button disappears when you edit the title. For now I've placed it to the left of the title which works ok, but if you or @shantanuraj have any better ideas? :hugs: 

image

rylek90:
This is not my final review, just some thoughts/answers. Code looks good, though 👍 

> @rylek90 in the end I found it was easier to have the button display as part of the page, instead of as part of the header :(
> 

Yes, this was my initial idea as well but I thought it used to be in header so I decided to go with more difficult way... which apparently led us nowhere. So all is good for me 🙌 I'll give this PR another go tomorrow morning, too tired for careful testing atm. Sorry for the delay 😅 

> Since this works for both the main page and the popup, I've replaced the popup specific attachment with this one.

Perfecto 💖 

> On positioning - I wanted to make the button appear after the text like usual for the minimal button, but I ran into issues when the page title has multiple lines. I tried putting the button itself inside the title div, which makes it appear just after the title text, but then the button disappears when you edit the title. For now I've placed it to the left of the title which works ok, but if you or Shantanu (edited to prevent unnecessary ping) have any better ideas? 

Unfortunately not, everything seems too dynamic in Notion and we already have ugly selectors here 😢 I'll try to figure something out with more clear head.

IMHO in popup view it looked good (especially with your styling changes 💅)... so maybe let's try to keep it there and display on left for main page? 🤔

lise-toggl:
@rylek90 sorry about the delay :see_no_evil: 

agree the topbar view looks much nicer, do you think it would be confusing for people to have the button in two different places or it's ok?

self note for future reference - these are the selectors I've looked at but don't reliably re-render on page change
- `.notion-frame`
- `.notion-topbar`
  - left side div (breadcrumbs)
  - last element in left side div (page name)
  - right side menu

and when they do change, starting a timer, changing the page and coming back - the running timer isn't recognised, it just shows the default "Start timer" link :(

tcrammond:
> and when they do change, starting a timer, changing the page and coming back - the running timer isn't recognised, it just shows the default "Start timer" link :(

you might be able to select the description fresh every time in `getDescription`, the old reference could be stale/remove now or something like that

lise-toggl:
@tcrammond sorry I didn't explain the issue properly - the description does get updated correctly when you switch pages.

The problem is that when you start a timer, leave the page then return, ideally we should show that there's a running timer for that page already. This doesn't happen - it always says "Start timer" even if you have a previously running timer from that page.

I've added back @rylek90's code for the header and the current state is:
- On main pages, show the button icon to the left of the page title
- On popup pages, show a styled button in the popup header

This is not ideal but at least we have buttons on both page types :smile:

PR #1624, opened 2019-12-16
Add Fallback login

### LINKED ISSUE #1618: Add Fallback login option
### Describe the new feature

- Add Fallback login option to the login page in case of a login error in standard flow

### If we couldn't add this feature, is there a compromise you can think of?

- No, related to several cases mentioned in #1495

shantanuraj:

- Adds fallback login option in case regular login fails

- Log out from button and webapp completely
- Fallback login will only show up in case regular login fails
- Use attached diff to force login view navigate to `chrome-extension://EXT_ID/html/login.html?source=quick-start`
- Try logging in using fallback form button should still work

```diff
diff --git a/src/scripts/components/LoginPage.tsx b/src/scripts/components/LoginPage.tsx
index 40858f0..bf8143e 100644
--- a/src/scripts/components/LoginPage.tsx
+++ b/src/scripts/components/LoginPage.tsx
@@ -52,7 +52,7 @@ async function login (setState: React.Dispatch<React.SetStateAction<LoginState>>
 }
 
 export default function LoginPage ({ source, isLoggedIn, isPopup }: LoginProps) {
-  const [ state, setState ] = React.useState<LoginState>({ loading: false, error: null, loggedIn: isLoggedIn });
+  const [ state, setState ] = React.useState<LoginState>({ loading: false, error: 'Random error', loggedIn: isLoggedIn });
   const { loading, error, loggedIn } = state;
 
   React.useEffect(() => {
```

Closes #1618

### DIFF
--- a/src/scripts/components/LoginForm.tsx
+++ b/src/scripts/components/LoginForm.tsx
@@ -27,7 +27,10 @@ export default function LoginForm ({ onSubmit, onSuccess, onError }: LoginFormPr
       const response = await onSubmit(email, password);
       setLoading(false);
 
-      if (!response) return;
+      if (!response) {
+        onError('Unknown Error');
+        return;
+      }
 
       if (response.success) {
         onSuccess();
--- a/src/scripts/components/LoginForm.tsx
+++ b/src/scripts/components/LoginForm.tsx
@@ -1,0 +1,124 @@
+import * as React from 'react';
+import styled from '@emotion/styled';
+
+import { Button } from '../@toggl/ui/components';
+import Spinner from './Spinner';
+
+interface LoginFormProps {
+  onSubmit: (email: string, password: string) => Promise<ButtonResponse | null>;
+  onSuccess: () => void;
+  onError: (error: string) => void;
+}
+
+export default function LoginForm ({ onSubmit, onSuccess, onError }: LoginFormProps) {
+  const [loading, setLoading] = React.useState(false);
+  const emailRef = React.useRef<HTMLInputElement>(null);
+  const passwordRef = React.useRef<HTMLInputElement>(null);
+
+  const handleSubmit = React.useCallback(
+    async (event: React.FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
+      const email = emailRef.current && emailRef.current.value;
+      const password = passwordRef.current && passwordRef.current.value;
+
+      if (!email || !password) return;
+
+      setLoading(true);
+      const response = await onSubmit(email, password);
+      setLoading(false);
+
+      if (!response) return;
+
+      if (response.success) {
+        onSuccess();
+      } else {
+        onError(response.error);
+      }
+    },
+    [emailRef, passwordRef, onSubmit, setLoading]
+  );
+
+  return (
+    <Form onSubmit={handleSubmit}>
+      <Field>
+        <Label htmlFor="login-email">Email Address</Label>
+        <Input
+          ref={emailRef}
+          id="login-email"
+          name="email"
+          type="email"
+          autoCorrect="off"
+          autoCapitalize="off"
+          autoComplete="username"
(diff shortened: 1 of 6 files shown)


src/scripts/components/LoginForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+      const email = emailRef.current && emailRef.current.value;
+      const password = passwordRef.current && passwordRef.current.value;
+
+      if (!email || !password) return;
+
+      setLoading(true);
+      const response = await onSubmit(email, password);
+      setLoading(false);
+
+      if (!response) return;
```
rylek90: Just curious face: shouldn't we handle it at least somehow?

shantanuraj: Good catch, it's an impossible case in this flow, but still added `Unkown Error` state in accordance with the LoginPage's login handler

src/scripts/components/LoginPage.tsx
```diff
@@ -91,22 +109,25 @@ export default function LoginPage ({ source, isLoggedIn, isPopup }: LoginProps)
...
-          </Subheading>
           <Row>
             <a href="login.html?source=install" style={{ marginBottom: 21 }}>
               <Button>Try again</Button>
             </a>
-            <Subheading>or</Subheading>
-            <a href="mailto:support@toggl.com" style={{ color: 'white' }}>
-              <Subheading style={{ margin: 0 }}>Contact support</Subheading>
-            </a>
+            <Subheading>or login directly</Subheading>
```
tcrammond:
```suggestion
            <Subheading>or log in directly</Subheading>
```

src/scripts/components/LoginPage.tsx
```diff
@@ -194,6 +215,14 @@ function SignupButton ({ isPopup }: Pick<LoginProps, 'isPopup'>) {
 
 const openPage = (url: string) => () => browser.tabs.create({ url });
 
+const performLogin = (username: string, password: string) => (
+  sendMessage({ type: 'login', username, password })
+);
+
+const onLoginSuccess = () => {
+  location.href = 'login.html?source=quick-start';
```
tcrammond: when using the secondary login, we seem to do both of these:
- Direct to the quick-start page
- Open the settings page in a new tab

we should only do the first one?

src/scripts/components/LoginPage.tsx
```diff
@@ -56,6 +58,13 @@ export default function LoginPage ({ source, isLoggedIn, isPopup }: LoginProps)
     if (source === 'quick-start' || loggedIn) document.title = 'Welcome to Toggl Button';
   }, [source, loading, error, loggedIn]);
 
+  const onLoginError = React.useCallback(
+    (error: string) => {
+      setState({ loading: false, loggedIn: false, error });
```
tcrammond: Should bugsnag be notified or does that get handled by existing stuff?

in any case, I guess we want to know about both original failures and then if the secondary form continues to fail 😬

shantanuraj: This existing effect should report the error

https://github.com/toggl/toggl-button/blob/2b8b1da4d5ad986b422dc01f88b3446c199f0e3f/src/scripts/components/LoginPage.tsx#L58-L62

src/scripts/components/LoginPage.tsx
shantanuraj: I couldn't reproduce this,
logging in opened the quick-start page for me
https://toggl.slack.com/files/U8U0D02S3/FRSDW6QRK/screen_recording_2019-12-17_at_12.14.31_pm.mov

src/scripts/components/LoginPage.tsx
shantanuraj: Fixed :bow:

PR #1626, opened 2019-12-17
feat(intercom): Add button to KB article edit page

lise-toggl:

Adds a button to intercom's article edit page:
image

Go to intercom's articles section, open an article and look for the button.

Requested internally

### DIFF
--- a/src/scripts/content/intercom.js
+++ b/src/scripts/content/intercom.js
@@ -31,3 +31,21 @@ togglbutton.render('.conversation__card__content-expanded__controls .inbox__conv
 
     elem.appendChild(link);
   });
+
+togglbutton.render('.articles__editor__header-text:not(.toggl)',
+  { observe: true },
+  function (elem) {
+    const descriptionSelector = () => {
+      const description = elem.textContent;
+      return description ? description.trim() : '';
+    };
+
+    const link = togglbutton.createTimerLink({
+      className: 'intercom',
+      description: descriptionSelector,
+      buttonType: 'minimal'
+    });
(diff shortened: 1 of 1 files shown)


src/scripts/content/intercom.js
```diff
@@ -31,3 +31,21 @@ togglbutton.render('.conversation__card__content-expanded__controls .inbox__conv
...
+      const description = elem.textContent;
+      return description ? description.trim() : '';
+    };
+
+    const link = togglbutton.createTimerLink({
+      className: 'intercom',
+      description: descriptionSelector,
+      buttonType: 'minimal'
+    });
+    link.style.margin = '3px 15px';
```
rylek90: Can't we include it in css instead? 🙏

shantanuraj: I would refrain from adding more new styles to that common CSS file,
we should ideally allow CSS files for the content scripts and have them live next to the script itself,
having styles in content scripts seems like an acceptable alternative until we move towards independent CSS files.

The monolith CSS file is difficult to reason about and is easy to miss, when refactoring/adding/removing integrations

lise-toggl: @rylek90 same as :point_up:, I didn't want to add another single use class in there. Are you ok with keeping it in the script?

rylek90: > we should ideally allow CSS files for the content scripts and have them live next to the script itself

This sounds great 🙌 

> Are you ok with keeping it in the script?

Yeah, sure 😅

Issue #1623, opened 2019-12-16
Jira integration: cannot select project from modal view


mrjones2014:
* **OS version**: Windows 10
* **Browser version**: Chrome 79
* **Extension version**: Version 79.0.3945.79 (Official Build) (64-bit)

### 🐛 Describe the bug

the Project dropdown is no longer clickable when you click "Start Timer". Also the "Start Timer" button appears twice.

### Expected behaviour

There should only be 1 "Start Timer" button and the "Project" dropdown should be clickable.

### Steps to reproduce

1. Enable the Jira integration
2. Go to a board for a sprint
3. Open a Jira story modal view
4. Click "Start Timer" on the Jira story
5. Attempt to set the Project. It won't work.

### Other details or context

It _does_ work if you go to the full page for the Jira story, but not the modal view.

lise-toggl:
@mrjones2014 thank you for the bug report!

We are working on a fix for the project dropdown at the moment. Could you confirm if both the project and tag dropdowns don't work for you, or just the project one?

Re the timer appearing twice, are you comfortable saving or taking a screenshot of the relevant DOM portion and sending it into our support team?

mrjones2014:
@lise-toggl confirmed, the issue is for both dropdowns.

Here's some screenshots of the DOM with sensitive information censored.

![image](https://user-images.githubusercontent.com/8648891/71669404-8ae77880-2d3a-11ea-9f23-67cb63bc22c5.png)


![image](https://user-images.githubusercontent.com/8648891/71669468-c08c6180-2d3a-11ea-9562-f3b5b7cef4d1.png)

mrjones2014:
If it helps, the URL I am using if of the format myorgname.atlassian.net

lise-toggl:
Thanks @mrjones2014! We're trying some ideas for the dropdown issue and hopefully will find a solution soon.

For the double button issue, would you be able to show me any ancestor DOM elements with the `.toggl` class? Thanks :)

mrjones2014:
@lise-toggl I think it may be on Jira stories that are part of an Epic, because it shows 2 Jira numbers (the Epic number / the Story number)

![image](https://user-images.githubusercontent.com/8648891/71976479-6aa93500-31e4-11ea-90eb-319b96771f50.png)

lise-toggl:
@mrjones2014 awesome - it was indeed the epic and we've been able to reproduce it now. Thanks for the screenshots, I should be able to fix it soon :)

[...]

Issue #1622, opened 2019-12-16
TRELLO card gets closed when project dropdown is clicked


teusch:
Please follow this guide when creating a new issue. This speeds up the process of replicating the issue and finding a solution.

### Environment details

* Operation system: OS Name	Microsoft Windows 10 Enterprise

* Extension version: 1.52.0

### Expected behavior
when I open the toggl button popup on the Trello card and I try to choose a project I expect the the trello card to stay open. 

### Actual behavior
when I open the toggl button popup on the Trello card and I try to choose a project the trello card closes 
### Steps to reproduce the behavior
go to trello 
open a card 
click "start Timer"
click on project dropdown
### Additional details
Happens in chrome and firefox
also on colleague computer
* screenshots or screencapture
![image](https://user-images.githubusercontent.com/5502583/70896414-f3bbba80-1ff0-11ea-8286-066ca0f378c8.png)

Issue #1448, opened 2019-07-04
Jira Kanban board - can't filter projects or tags


lise-toggl:
* **OS version**: W10
* **Browser version**: Chrome 75
* **Extension version**: 1.37.4

#### Relevant integration (if any):

JIRA

### 🐛 Describe the bug

When starting a timer in JIRA from the Kanban board, the "Find project" and "Find tags" fields aren't selectable, so you have to scroll through the list to find your project/tag.
![image](https://user-images.githubusercontent.com/50156618/60642941-d21cab80-9e74-11e9-96cb-c25417aefc4f.png)

The field works ok from the Issues page.

### Expected behaviour

The search field should be selectable so you can search for a project or tag.

### Steps to reproduce

1. Log into Jira
2. Open a Kanban board
3. Click on an Issue
4. Click Start timer
5. Open the projects dropdown
6. Click on the search field

### Other details or context

[case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/3155115/conversations/22433077780)

PR #1637, opened 2020-01-07
[WIP] Use iframe for popup edit form

lise-toggl:

Uses an iframe to display the popup edit form, to avoid weird interactions with underlying page.

Currently working:
- Creates an iframe when a timer is started
- Frame reuses edit view from the normal browser popup
- Frame is deleted on click outside

Not working:
- Firefox
- Frame should be deleted on pressing "Enter" inside the popup
- Autocomplete boxes should be a little taller

:grimacing: 

If this works it would fix a bunch of issues, e.g. #1623, #1622, #1448, and hopefully be more hardy in general :wrench:

### DIFF
--- a/src/chrome-manifest.json
+++ b/src/chrome-manifest.json
@@ -54,5 +54,5 @@
   },
   "content_security_policy":
     "script-src 'self'; object-src 'self';",
-  "web_accessible_resources": ["html/login.html"]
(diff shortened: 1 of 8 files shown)


lise-toggl:
Closing as changing the DOM position per Tyler also works and is a much easier fix for our current issues :tada:

PR #1639, opened 2020-01-08
GitLab: Add GitLab labels as Toggl Tags

glensc:
Please remember the [Contributing Guidelines](https://github.com/toggl/toggl-button/blob/master/.github/CONTRIBUTING.md) :heart:

## :star2: What does this PR do?

GitLab: Add GitLab labels as Toggl Tags. Requires GitLab 12.7

## :bug: Recommendations for testing

All changes should be tested across Chrome and Firefox.

## :memo: Links to relevant issues or information

- https://github.com/toggl/toggl-button/issues/1657

PR #1642, opened 2020-01-09
fix(autocomplete): Fix escaping of tag names in html title attribute

glensc:
There's bug if tag containing `::` is added, autocomplete code gets fatal error because `::` indicates DOM selector, not verbatim text.

This was extracted out of https://github.com/toggl/toggl-button/pull/1639

Docs:
- https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -1418,8 +1418,8 @@ window.TogglButton = {
       .map(t => `
         <li
           class="tag-item"
-          data-wid=${escapeHtml(t.wid)}
-          title=${escapeHtml(t.name)}
+          data-wid="${escapeHtml(t.wid)}"
+          title="${escapeHtml(t.name)}"
         >
         ${escapeHtml(t.name)}
         </li>
--- a/src/scripts/lib/autocomplete.js
+++ b/src/scripts/lib/autocomplete.js
@@ -708,17 +708,31 @@ TagAutoComplete.prototype.selectTag = function (e) {
 };
 
 TagAutoComplete.prototype.setSelected = function (tags) {
-  let i; let item;
-
   this.clearSelectedTags();
 
+  /**
+   * Create map of elements keyed to element .title attribute value.
+   * @param {NodeList} elements
+   */
+  const mapByTitle = function (elements) {
+    let i;
+    let title;
+    const result = {};
+
+    for (i = 0; i < elements.length; i += 1) {
+      title = elements[i].getAttribute('title');
+      result[title] = elements[i];
+    }
+
+    return result;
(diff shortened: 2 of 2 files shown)


lise-toggl:
@glensc thank you for the PR! Could you please repro steps? I tried a few things with tags containing `::` but couldn't recreate the error yet. Thanks!

glensc:
@lise-toggl As the changeset was extracted out of #1639, the reproducer is to add tags containing `::`, perhaps something:

```js
 const link = togglbutton.createTimerLink({
/*...*/
      tags: ["tag::colon", "namespace::value"],
/*...*/
    });
```

I also noticed identical error when I was (mistakenly) trying to add a tag with a newline (`\n`). should those be escaped as well?

glensc:
https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

I need to escape space as well. so I'm thinking different approach, instead of escaping, grab all tags and do getAttribute comparison? as I can't possibly know what needs to be escaped. for example `\n` needs to escaped as well, even it's not mentioned in `querySelector` documentation.

glensc:
@lise-toggl you can test something like this:

```diff
diff --git a/src/scripts/content/gitlab.js b/src/scripts/content/gitlab.js
index 41dbdbc2..2524e231 100644
--- a/src/scripts/content/gitlab.js
+++ b/src/scripts/content/gitlab.js
@@ -29,6 +29,7 @@ togglbutton.render(
     const link = togglbutton.createTimerLink({
       className: 'gitlab',
       description: description,
+      tags: ['lol::cats', 'lol space', 'lol \\backspace', 'newline \n escaped'],
       projectName: projectElem.textContent
     });

@@ -66,6 +67,7 @@ togglbutton.render(
     const link = togglbutton.createTimerLink({
       className: 'gitlab',
       description: description,
+      tags: ['lol::cats', 'lol space', 'lol \\backspace', 'newline \n escaped'],
       projectName: projectElem.textContent
     });
```

glensc:
also, looks like the .title attribute is broken after space, i.e value is prematurely terminated when space is encountered. i.e tags that came from toggl itself.

image

glensc:
I've pushed new logic, that does not involve escaping: 85cc5ebf

glensc:
Ok, the HTML was not properly escaped: a5bfd3e7. I tagged the fix as security.

Do you need for me to extract that fix as separate PR?

cc @tcrammond @lise-toggl

tcrammond:
No, it's fine here, thank you! Will review the updated PR soon.

PR #1652, opened 2020-01-15
Rearrange edit form buttons

### LINKED ISSUE #1632: Change Delete Button
### Describe the new feature

Hi there, just wanted to send some feedback on the new browser plugin UI. It’s looking great, and seems much better than before - especially grouping entries by date, which makes things much easier to see.
However, when entering things quickly, I keep clicking the delete button instead of the Done button.
The hit area for delete is double that of both Done and Cancel (due to its width), and it sits very close to the Done button.
For quick entry, I think it would be helpful to make the delete button smaller, and move it away from the “Done” button.

image

[Case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/534741/conversations/25073053082)

lise-toggl:

- Make the "Done" button 100% wide
- Move the "Cancel" button next to the "Delete" button

This makes the popup's edit form look more like the integration popup (where "Done" is 100% wide) and groups the less commonly used "Cancel" and "Delete" options together.

image

Buttons should still work as before.

Closes #1632

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -147,11 +147,11 @@ window.TogglButton = {
     '</div>' +
     '</div>' +
     `
-      <div id="tb-actions-button-group">
-        <button type="button" id="tb-edit-form-cancel" tabindex="0" class="TB__Secondary__button">Cancel</button>
+      <div class="tb-actions-button-group">
         <div id="toggl-button-update" tabindex="0" class="TB__Button__button">Done</div>
       </div>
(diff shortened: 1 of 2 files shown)


lise-toggl:
For comparison, the current arrangement:
image

lise-toggl:
Can you try reloading the extension? I had to make some changes to `background.js` which doesn't get picked up until it's reloaded, I think it's that :(

tcrammond:
yeah ^ that's probably it

PR #1661, opened 2020-01-28
style(brokerengine): Update hover styles for toggl buttons on task queue

christianztamayo:

This PR updates the hover styling for toggl buttons on BrokerEngine's task queue view.

Lowered opacity on buttons that are neither being hovered nor active.

Hover:
Screen Shot 2020-01-28 at 5 03 39 PM

Active:
Screen Shot 2020-01-28 at 5 04 15 PM

### DIFF
--- a/src/styles/style.css
+++ b/src/styles/style.css
@@ -1500,13 +1500,11 @@ body.notion-body.dark .toggl-button.notion {
 }
 
 [data-toggl="taskRow"] .toggl-button.brokerengine:not(.active) {
-  opacity: 0;
-  visibility: hidden;
+  opacity: 0.4;
 }
 
 [data-toggl="taskRow"]:hover .toggl-button.brokerengine {
   opacity: 1;
-  visibility: visible;
 }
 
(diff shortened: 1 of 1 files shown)


tcrammond:
Hey, thanks for submitting your PR. Unfortunately this change to the buttons in the task list is probably not something we'll want to approve. It won't match our style guide or how the other major integrations function

The Toggl Button [style guide](https://github.com/toggl/toggl-button/blob/master/.github/CONTRIBUTING.md#adding-new-integrations) (see "For integrations that show many buttons in a list") suggests that the buttons should be hidden by default in lists of tasks. We've heard feedback in the past that the UI gets too busy with Toggl buttons everywhere, and we don't really want to make exceptions for it at the moment 😞 

Do you have feedback to share about why this change would be better for you?

christianztamayo:
Hey @tcrammond, apologies for the late response. I read the style guide actually, and was taking my chances of getting this PR approved. This is mostly for improving the user experience, as per feedback from our users. Having it show on hover seems like an additional effort. Hence to conform, we opted to lowering the opacity when not hovering. Would it help if we lower it further?

PR #1666, opened 2020-02-05
fix(redmine) Prevent positioning in sidebar

jrenggli:

Currently the button is rendered in the first `h2` tag found. Depending on installed plugins in Redmine there are multiple `h2` tags present. This leads to having the button rendered at the wrong position and using the wrong title.

This PR makes the selectors more specific by using `#content h2` to detect `numElem` and button position.

Use Redmine with Plugins installed. e.g. [Helpdesk](https://www.redmineup.com/pages/de/plugins/helpdesk)

### DIFF
--- a/src/scripts/content/redmine.js
+++ b/src/scripts/content/redmine.js
@@ -1,10 +1,10 @@
 'use strict';
 
 togglbutton.render(
-  'body.controller-issues.action-show h2:not(.toggl)',
+  'body.controller-issues.action-show #content h2:not(.toggl)',
   {},
   function (elem) {
-    const numElem = $('h2');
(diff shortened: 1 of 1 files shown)


glensc:
Oh wow, so old commit.

```
From: Jonas Renggli <jonas.renggli@visol.ch>
Date: Tue, 14 Aug 2018 15:21:59 +0200
Subject: [PATCH] fix(redmine) Prevent positioning in sidebar
```

https://github.com/toggl/toggl-button/pull/1666.patch

I got confused for a second, that I'm looking at outdated GitHub repository, or what is going on 🤣 

image

PR #1674, opened 2020-02-13
Remove "checkPermissions" logic on load/login

### LINKED ISSUE #1663: Opening Chrome causes Toggl button settings to open on every load
* **Windows 10**: 
* **Chrome awaiting more information from user**: 
* Windows v7.4.1112

### 🐛 Describe the bug

The case [here](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2097915/conversations/25617417587), is the second case I've seen where a user reports that the Toggl button settings menu opens each time they open Chrome. 

### Expected behaviour

Opening Chrome shouldn't open the Toggl button settings menu

### Steps to reproduce

I cannot reproduce, but the users involved open Chrome, the Toggl button settings open immediately each time.

tcrammond:

Remove some `checkPermissions` logic that would automatically open the login page if it thinks you don't have any permissions enabled, on load/login (with login meaning, any time user is fetched, I think).

Do not think this is needed any more with the new onboarding page(s).

Everything seems sound, and we're not missing out on anything as a result of this change.

 Closes #1663.

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -1192,7 +1192,6 @@ window.TogglButton = {
         onLoad: function (xhr) {
           if (xhr.status === 200) {
             TogglButton.setupToken(xhr.responseText);
-            TogglButton.queue.push(TogglButton.checkPermissions);
             TogglButton.fetchUser()
               .then((response) => {
                 TogglButton.refreshPage();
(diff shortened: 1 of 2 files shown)


src/scripts/background.js
```diff
@@ -2254,22 +2253,6 @@ window.TogglButton = {
     }
   },
 
-  checkPermissions: async function (show) {
-    const dontShowPermissions = await db.get('dont-show-permissions');
```
shantanuraj: This PR is good to go!

There's some additional code we can get rid of what do you think?

image

tcrammond: Pushed a commit 🙆‍♂

PR #1682, opened 2020-02-25
Eventum: Add support for fetching custom fields as tags

glensc:

Similar integration as to GitLab: https://github.com/toggl/toggl-button/pull/1639

This is probably not useful on its own, but pre-requisite for https://github.com/toggl/toggl-button/pull/1683


glensc:
@tcrammond ping

PR #1715, opened 2020-04-17
Add notification when trying to stop entry with unmet constraints

### LINKED ISSUE #1430: Can't stop TE with unmet constraints
### 🐛 Describe the bug
If you try to stop a TE that has unmet constraints, visually nothing happens. There's a failed `PUT` request that explains what needs to be done.

We need to surface that error so the user is aware of what's going on
### Expected behaviour
A notification/toast should popup explaining why the TE can't be stopped/created.

### Steps to reproduce
* Have workspace with TE constraints
* Have TE with unmet constraints
* Try to stop that TE

rylek90:

For now it only prints error from backend, which is descriptive enough, imho. We might actually add some actions pointing to ws settings, or something.

⚠️ All changes should be tested across Chrome and Firefox. ⚠️ 

* Set some constraints in ws settings
* Try to stop Time Entry in button without ☝️ fields filled in
*  There should be a notification 

Try pomodoro with "stop time entry" option set too

Closes #1430

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -874,6 +874,9 @@ window.TogglButton = {
                   browser.tabs.sendMessage(tabs[0].id, { type: 'stop-entry', user: TogglButton.$user });
                 }));
             }
+            if (xhr.status === 400 && xhr.response && xhr.response.includes('please add a')) {
+              TogglButton.showUnmetConstraintsNotification(xhr.response);
+            }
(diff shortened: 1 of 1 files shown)


tcrammond:
Good to go? 🙏

rylek90:
> Good to go? 🙏

I guess it should be 😂

PR #1762, opened 2020-06-22
feat(asana): Add Automatic Tagging feature to Asana

tcrammond:
Please remember the [Contributing Guidelines](https://github.com/toggl/toggl-button/blob/master/.github/CONTRIBUTING.md) :heart:

## :star2: What does this PR do?

Adds automatic tagging support to Asana.

## :bug: Recommendations for testing

All changes should be tested across Chrome and Firefox.

Verify creating entries with / without automatic tagging in these views:
* My Tasks https://app.asana.com/0/729616262525591/list
* My Tasks, detail view https://app.asana.com/0/729616262525591/1181404311905337
* Spreadsheet https://app.asana.com/0/729617904817453/list
* Spreadsheet detail view
* Board detail view https://app.asana.com/0/729617904817453/960516404216916

## :memo: Links to relevant issues or information

Closes #1691.

PR #1766, opened 2020-06-23
fix(asana): Update project selector for My Tasks page

### LINKED ISSUE #1757: TB is not matching projects when used in Asana [My Tasks view]
* **OS version**: Windows 10
* **Browser version**: 83.0.4103.106
* **Extension version**: 1.62.1.

Asana [My Tasks view]

### 🐛 Describe the bug

When using TB in Asana on My Tasks view, the project is not being picked up even though matching project exists in Toggl. However, the project is matched if using the Toggl button from Task details

### Expected behaviour

Project should be automatically chosen if the exact ones exist in Toggl.

### Steps to reproduce

1. Create a project with the same name in Asana and Toggl
2. Go to Asana and click on My Tasks on the left
3. Create TE for that task using TB in Asana
4. It will not pick project.

Screenshot 2020-06-18 at 15 32 15

If you click on a Task and click on TB from there, the Project will match automatically:

Screenshot 2020-06-18 at 16 16 43

### Other details or context

[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/2759100/conversations/24705809818)

lise-toggl:

- Project names will be picked up again on the My Tasks page
- Move button to far right so they align if a project is present

1. Go to My Tasks page
2. Time a task with a project
3. Project is now picked up
4. Icons no longer do this
image

Closes #1757

### DIFF
--- a/src/scripts/content/asana.js
+++ b/src/scripts/content/asana.js
@@ -70,30 +70,49 @@ togglbutton.render('.SpreadsheetRow .SpreadsheetTaskName:not(.toggl)', { observe
   }
 );
 
-// 2019 My Tasks view, possibly other similar views.
+// 2020 My Tasks view, possibly other similar views.
 togglbutton.render('.MyTasksTaskRow:not(.toggl)', { observe: true },
   function (elem) {
     if (elem.querySelector('.toggl-button')) {
       // Due to the way this UI is rendered, we must check for existence of old buttons manually.
       return;
     }
-    const container = elem.querySelector('.ItemRowTwoColumnStructure-left');
     const descriptionSelector = () => elem.querySelector('.TaskName textarea').textContent;
 
-    const projectSelector = () => {
-      const projectElement = elem.querySelector('.TaskRow-pots .Pill');
+    // attempt at separating projects and tags, which are not differentiated in the dom
+    // assume first pill is a project and any others are tags
+    // misses tags which are in the "..." overflow, and if there is a tag without a project
+    const pillSelector = (type) => {
+      const pills = [...elem.querySelectorAll('.NavigationLink')]
+        .map(pill => pill.textContent.trim());
+      if (type === 'project') {
+        return pills.length ? pills[0] : '';
+      } else if (type === 'tags') {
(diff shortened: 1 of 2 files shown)


src/scripts/content/asana.js
```diff
@@ -70,18 +70,17 @@ togglbutton.render('.SpreadsheetRow .SpreadsheetTaskName:not(.toggl)', { observe
...
     if (elem.querySelector('.toggl-button')) {
       // Due to the way this UI is rendered, we must check for existence of old buttons manually.
       return;
     }
-    const container = elem.querySelector('.ItemRowTwoColumnStructure-left');
     const descriptionSelector = () => elem.querySelector('.TaskName textarea').textContent;
 
     const projectSelector = () => {
-      const projectElement = elem.querySelector('.TaskRow-pots .Pill');
+      const projectElement = elem.querySelector('.PotPillsContainer');
```
tcrammond, line 83: This selector doesn't quite work, it will also pick up any tags that are present (breaking the project matching) 😞 since they're all shown as a pill in the same container

Screenshot 2020-06-23 at 11 06 50

actually, I think I've worked on the same bug in the other Asana PR here 🙈  sorry I didn't realise it was an existing issue, just happened to notice it https://github.com/toggl/toggl-button/pull/1762/files#diff-50b20c2c72dfe8169e6d571835144c22R91 

do you want to bring that selector ^ in here (if you test that it works)? since this PR still has some other fixes in it we may as well do it independently of the bigger PR that adds tagging support.

lise-toggl: Damn :see_no_evil: 

That selector doesn't match anything for me at all, I wonder if there are multiple Asana views out again? :thinking: 

Here's what it looks like on mine - a red project followed by a red tag - they're identical :sob: 
image

Only thing I can think of is to assume the first pill is a project, and the rest are tags?

tcrammond: ^ I think that may be the best course here; will need to test all views with it though

lise-toggl: Updated!

I think the My Tasks page is the only one with this problem, the other views do show tags differently to the project.

PR #1769, opened 2020-06-25
feat(dark-mode): Add dark mode

### LINKED ISSUE #1575: [Phoenix] Dark Mode support
- Add dark mode support

lise-toggl:

- Add a dark mode toggle on the settings page
- When enabled, make the settings page, browser popup and integration popups dark
- Update the timer block slightly

image

image

image

- Turn it on and click around, is there anything that looks weird?

Closes #1575

### DIFF
--- a/src/html/settings.html
+++ b/src/html/settings.html
@@ -3,6 +3,8 @@
   <head>
     <title>Toggl Button - Settings</title>
     <link rel="stylesheet" href="../styles/settings.css"/>
+
+    <script src="../scripts/settings-theme.js"></script>
   </head>
   <body data-show-review-banner="false">
     <nav>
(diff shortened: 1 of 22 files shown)


src/scripts/components/BillableIcon.tsx
```diff
@@ -4,26 +4,11 @@ import styled from '@emotion/styled'
 
 import * as color from '../@toggl/style/lib/color';
 
-interface Theme {
```
lise-toggl, line 7: Was this the start of a react theme function? I don't think it was being used so removed it :eyes:

src/scripts/components/BillableIcon.tsx
```diff
@@ -4,26 +4,11 @@ import styled from '@emotion/styled'
 
 import * as color from '../@toggl/style/lib/color';
```
lise-toggl, line 5: Someone should change usages of this to the new css variables, at some point

src/styles/style.css
```diff
@@ -1,6 +1,29 @@
...
+  --lighter-grey: #e3e4e4;
+  --grey: #363636;
+  --mineshaft: #323232;
+  --dark-grey: #303030;
+  --another-black: #222;
+  --darker-black: #1e1e1e;
+
+  --red: #d92b2b;
+  --red-on-dark: #ff5050;
+}
```
lise-toggl, line 15: importing these from `'./colours.css` did not work :question:

src/html/settings.html
```diff
@@ -221,6 +221,13 @@ <h3>Other settings</h3>
               <span class="description">Select text and start a timer with that description from the right-click menu. If nothing is selected, the page title is used as the description.</span>
             </label>
           </div>
+          <div class="checkbox field">
+            <input type="checkbox" id="dark_mode">
+            <label for="dark_mode">
+              <span class="setting">Enable dark mode</span>
+              <span class="description">Use dark mode for the Toggl Button popups.</span>
```
shantanuraj, line 228:
```suggestion
              <span class="description">Use dark mode</span>
```
It affects the settings page as well

src/scripts/components/BillableIcon.tsx
shantanuraj: I don't recall why this was added

src/styles/style.css
shantanuraj: Weird can't get it to work as well

src/scripts/background.js
```diff
@@ -682,6 +684,11 @@ window.TogglButton = {
     return timeEntries[timeEntries.length - 1];
   },
 
+  checkDarkMode: async function () {
+    const darkMode = await db.get('darkMode');
+    return darkMode;
```
shantanuraj, line 689:
```suggestion
    return darkMode ?? matchMedia('(prefers-color-scheme: dark)').matches;
```
Could we do this? So it defaults to dark-mode if user has enabled it in their system settings

dooart:
WHAT!!!!!!!!! DARK MODE! ?!?!?!?!?!? OMGOMGOMG

src/scripts/components/BillableIcon.tsx
tcrammond: I think it's copypaste from the web codebase

lise-toggl: Ok, sounds like it's safe to remove :hugs:

src/scripts/background.js
lise-toggl: :clap: good idea, `??` doesn't work in our current babel config so had to change it a little. I could not actually get this to work in testing though, does it work for you?

lise-toggl: Ok let's see if this works :thinking:

lise-toggl:
@tcrammond good catch, that flashing was horrible :frowning_face: 

I've reworked the links in the settings page to just swap the active class around instead of loading a new page. It still flashes on first load, there's probably something we can do to block rendering until the value is resolved right? Is it worth bothering with?

Updated the icons to stand out a little more :+1: 

I think it's better to let people define dark/light indepdendent of their system; mobile had a lot of people asking for exactly that and I expect we'd get the same (assuming people find out the feature exists :smile: )

tcrammond:
Nice 👏  

> It still flashes on first load, there's probably something we can do to block rendering until the value is resolved right?

yeah, might be straightforward 🤔

src/scripts/settings-theme.js
```diff
@@ -0,0 +1,2 @@
+const darkMode = JSON.parse(localStorage.getItem('darkMode'));
```
shantanuraj, line 1:
```suggestion
const darkMode = JSON.parse(localStorage.getItem('darkMode')) ?? window.matchMedia('(prefers-color-scheme: dark)').matches;
```
Should we default to system-theme?

lise-toggl: should flow through from here right? https://github.com/toggl/toggl-button/pull/1769/files#diff-e7905509b504970b523439849d059794R12

lise-toggl:
@tcrammond now also storing dark mode value in localStorage (as it's not async) and fetching it before load :eyes: 

we could put it inline in the settings file but that would require the inline script permission and i didn't want to do that :grimacing:

src/scripts/settings-theme.js
shantanuraj: Not for first load I think, but that's fine best not to duplicate code and make the logic more obscure.

lise-toggl:
oh damn :cry: : `window.matchMedia('(prefers-color-scheme: dark)').matches` didn't work for me either but i thought it was just another :penguin: thing

i tried reproducing the flash a bit today but it's not doing that for me in chrome or firefox, not sure :thinking: 

i hope someone else can pick this up, good luck guys :wave:

shantanuraj:
Pushed a commit to default to system-theme if no value is stored in `Db` solves the initial render flash for me.

tcrammond:
Tested a bit, works better now @shantanuraj 🙌 

I pushed a quick commit to fix a problem where: I would be running on the system dark preference, then toggle it to ON in settings, but the class would be removed - putting me in light mode . Please sanity check it

❓ After you toggle the setting to ON and then OFF, it will never use your system preference again because there is a non-null value for the setting. Should we pivot to `use system preference / light mode / dark mode` selection?

shantanuraj:
Opened new issue #1771 to tackle system-default theme preference. If you give it a 💚  I'll merge this one.

nmiquan:
Could someone increase the contrast when hovering in the project dropdown list? When I'm choosing project in the projects list, I couldn't see which one I am selecting.

shantanuraj:
Hi @nmiquan I've created an issue at #1813 the team will look into it

ryan-efendy:
@shantanuraj can we have the dark mode be enabled automatically? via system color scheme
i.e. activate when system dark mode is on and vice versa when dark mode is off?

Issue #1774, opened 2020-07-08
Button not showing in webapp Todoist


nidarasheed:
* **OS version**: 
* **Browser version**: 
* **Extension version**: 

#### Relevant integration (if any):

Todoist
### 🐛 Describe the bug

After a recent update, the button isn't showing in Todoist. It's there, we can click on it, but it won't show. 

![image](https://user-images.githubusercontent.com/49634673/86890265-7aafbc00-c116-11ea-8ec8-037bfe2e94a6.png)

![image](https://user-images.githubusercontent.com/49634673/86890322-9024e600-c116-11ea-8f58-a5af01a173e8.png)

### Expected behaviour

Button should show on hover. 
### Steps to reproduce

Steps to reproduce the behaviour:
1. Go to todoist
2. Open any task list
3. Hover where TB should be. 

### Other details or context

[Case Reference 1
](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/search/conversations/106481300026004)

[Case Reference 2
](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/search/conversations/106481300025950)

[Case Reference 3](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/search/conversations/106481300025965)

nidarasheed:
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/534741/conversations/106481300026033)

nidarasheed:
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2144650/conversations/106481300026080)

edtamayo:
[Case Reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/3824737/conversations/106481300026172)

sKawashima:
console err

> Refused to load the image '<URL>' because it violates the following Content Security Policy directive: "default-src 'self' 'unsafe-eval' <URL> <URL> <URL> <URL> <URL> <URL> <URL> <URL> <URL> <URL> <URL> <URL> <URL> <URL> <URL> <URL>". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.

[...]

PR #1788, opened 2020-07-16
fix(todoist): Fix button not showing on Todoist

dpsxp:
Change button to SVG instead of using background images to avoid conflicts with CSP.  There are still more components using background images but they're only inside the popdown and it will not break the integration if the icon is missing. So I didn't change that to keep the scope of this very minimal.

- Open Todoist
- Check if the button is there
- Check if the integration is working as expected

#### Side effects ⚠️ 
Since we're changing the structure of the button, it may make it look bad in other integrations. The size still the same so unless we were doing some padding it should "just work".

#1774


src/scripts/common.js
```diff
@@ -1,4 +1,6 @@
 import { ProjectAutoComplete, TagAutoComplete } from './lib/autocomplete';
+/* eslint-disable-next-line import/no-webpack-loader-syntax */
+import togglButtonSVG from '!!raw-loader!./icons/toggl-button.svg';
```
dpsxp, line 3: Using the `raw-loader` here so we can still use the shared SVG instead of transforming it into a URL. I'm using the inline option because we may want to use it like the other icons later. Please let me know if this is a problem, I can try to set up something on webpack config to handle it.

shantanuraj: Makes sense to me 👍

rylek90: Yeah, it's fine. Transforming it into a URL wouldn't solve the problem with Todoist's CSP 💯

dpsxp:
Thanks @shantanuraj , as I mentioned in the description I was expecting some side effects but not like this 🙈

rylek90:
Hold on I've just found little quirk. Github integration missing text
image

PR #1792, opened 2020-07-20
style: Make continue button easier to click

seanbeaton:
This PR makes the clickable area for the continue button a little larger and fixes the mis-positioned "target" element due to the missing `position: relative;` on the parent.

Before, only the triangle of the play button was clickable. 

Now, a square area with ~5px paddings around the button is clickable. 

This doesn't interfere with the tags or billable icons that appear nearby.


seanbeaton:
Is there anything I should do to get this reviewed?

PR #1796, opened 2020-07-23
fix(zendesk): Fix missing TE description

### LINKED ISSUE #1794: Zedesk integration TE description issue
* **OS version**: Windows 10
* **Browser version**: Chrome 84.0.4147.89
* **Extension version**: 1.64.1

zendesk.com

### 🐛 Describe the bug

If you start a timer from the ticket header the description for the TE will be pulled from there, but, if you start it using the sidebar you get the following: 

image

image

### Expected behaviour

The description should be grabbed from the ticket header with both approaches 

[Intercom case 🔗 
](url
image
)

dpsxp:
Fix code to get the TE description based on the current ticket title. Also, a small improvement on the customer navigation button.

- Get a Zendesk account
- Open the ticket view
- Check if when starting the timer te description is the same as the ticket title

close #1794

### DIFF
--- a/src/scripts/content/zendesk.js
+++ b/src/scripts/content/zendesk.js
@@ -43,16 +43,19 @@ togglbutton.render('[data-test-id="customer-context-tab-navigation"]', { observe
   const titleFunc = function () {
     let description;
 
-    const titleElem = document.querySelector('[aria-label="Subject"]');
     const ticketNum = location.href.match(/tickets\/(\d+)/);
 
-    if (titleElem !== null) {
-      description = titleElem.value.trim();
+    if (ticketNum) {
+      const id = ticketNum[1].trim();
+      const titleElem = document.querySelector(`[data-side-conversations-anchor-id="${id}"] [data-test-id="ticket-pane-subject"]`);
+
+      if (titleElem !== null) {
+        description = titleElem.value.trim();
+      }
+
+      description = '#' + id + ' ' + description;
     }
 
-    if (ticketNum) {
-      description = '#' + ticketNum[1].trim() + ' ' + description;
-    }
     return description;
   };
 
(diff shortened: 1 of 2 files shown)


src/scripts/content/zendesk.js
```diff
@@ -43,7 +43,7 @@ togglbutton.render('[data-test-id="customer-context-tab-navigation"]', { observe
   const titleFunc = function () {
     let description;
 
-    const titleElem = document.querySelector('[aria-label="Subject"]');
+    const titleElem = document.querySelector('[data-test-id="ticket-pane-subject"]');
```
rylek90, line 46: `data-test-id` 😂 

Question is when do Zendesk team realise and strip these from production. 🤔

dpsxp: Wondering if isn't intentional, but this is the only unique selector that I found. I hope they keep it 🙈

rylek90: So far so good! Maybe they will never switch it off 🙊

rylek90:
> But even beside that, I noticed that the button is readily available on a ticket that is in process of being created, but its description is not picked when starting a timer. Only when viewing an already created ticket.

Good catch ❤️ 

This can be possibly fixed by changing `const titleElem = elem.querySelector(...)` to a function 🤔

dpsxp:
Found a bug, when you have more than one ticket open it will always get the values from the first ticket.

dpsxp:
> Found a bug, when you have more than one ticket open it will always get the values from the first ticket.

It should be fixed now.

> But even beside that, I noticed that the button is readily available on a ticket that is in process of being created, but its description is not picked when starting a timer. Only when viewing an already created ticket.

I'm not sure if this is really something that worth doing, we don't have the `id` of the ticket yet and the user is only creating the ticket that he may or may not start right after. Also, I was unable to find a good selector to get the title 🙈

Issue #1775, opened 2020-07-08
Switching on Asana integration doesn't "stick" 


nidarasheed:
* **OS version**: MacOS 10.15.4
* **Browser version**: Chrome Version 83.0.4103.116 (Official Build) (64-bit)
* **Extension version**: 1.63.0

#### Relevant integration (if any):

Asana

### 🐛 Describe the bug

Choosing Asana as integration does not stick. When you refresh the page, it is unselected again. 

### Expected behaviour
Integration selection should stick. 

### Steps to reproduce
1. Open Integrations tab
2. Choose Asana
3. Refresh 

### Other details or context

[Slack](https://toggl.slack.com/archives/C9LMUPD9T/p1594203735014500)
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/534741/conversations/106481300026065)

kornelijepetak:
This in fact happens with all of the following:

* asana.com
* breeze.pm
* brokerengine.com.au
* clubhouse.io
* codeable.io
* dom-integration
* khanacademy.org
* gqueues.com
* heflo.com
* helpscout.net
* huskymarketingplanner.com
* liquidplanner.com
* meistertask.com
* nozbe.com
* osticket.com
* sourcelair.com
* teamleader.eu
* vivifyscrum.com
* zenhub.com

PR #1802, opened 2020-07-27
fix(settings): Integrations settings not "checking"

dpsxp:

Fix integrations settings not showing the correct state for some domains

- Go to settings
- Integrations
- Check if when "Enable All" is clicked all the options are selected
- Check if when "Disable All" is clicked all the options are disabled

#1775

### DIFF
--- a/src/scripts/origins.js
+++ b/src/scripts/origins.js
@@ -12,7 +12,7 @@ export default {
     url: '*://*.any.do/*',
     name: 'Anydo'
   },
-  'asana.com': {
+  'app.asana.com': {
     url: '*://app.asana.com/*',
     name: 'Asana'
   },
@@ -62,11 +62,11 @@ export default {
     url: '*://*.bitrix24.com/*',
     name: 'Bitrix24'
   },
-  'breeze.pm': {
+  'app.breeze.pm': {
     url: '*://app.breeze.pm/*',
     name: 'Breeze'
   },
-  'brokerengine.com.au': {
+  'app.brokerengine.com.au': {
     url: '*://app.brokerengine.com.au/*',
     name: 'BrokerEngine'
   },
@@ -90,11 +90,11 @@ export default {
     url: '*://*.cloudes.me/*',
     name: 'Cloudes.me'
   },
-  'clubhouse.io': {
+  'app.clubhouse.io': {
     url: '*://app.clubhouse.io/*',
     name: 'Clubhouse'
   },
-  'codeable.io': {
+  'app.codeable.io': {
     url: '*://app.codeable.io/*',
     name: 'Codeable'
   },
@@ -232,7 +232,7 @@ export default {
     url: '*://keep.google.com/*',
     name: 'Google Keep'
   },
-  'khanacademy.org': {
+  'www.khanacademy.org': {
     url: '*://www.khanacademy.org/*',
     name: 'KhanAcademy'
   },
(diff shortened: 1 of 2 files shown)


src/scripts/settings.js
```diff
@@ -32,6 +32,14 @@ document.querySelector('#close-review-prompt').addEventListener('click', dismiss
   link.addEventListener('click', changeActiveTab);
 });
 
+const isNotTogglApp = (val) => {
+  if (!val) {
+    return true;
+  }
+
+  return (!!val.match(/toggl\./)) === false;
```
kornelijepetak, line 40: Also, I find this `!!` and `=== false` really detached from the semantics of what the code is doing, regardless of the fact that the output is correct.

I'd propose following the semantics of what the algorithm is actually doing, which is `whether there are any matches`:

`val.match(/toggl\./))?.length > 0`

src/scripts/origins.js
```diff
@@ -62,11 +62,11 @@ export default {
     url: '*://*.bitrix24.com/*',
     name: 'Bitrix24'
   },
-  'breeze.pm': {
+  'app.breeze.pm': {
```
kornelijepetak, line 65: I have mentioned all of these in the issue, but did you actually go thru all of them and checked whether their respective apps are actually under such subdomains?

Cause otherwise, I'm really confused as to why this solved the problem.

But partially I can't remember what was wrong for those domains in the first place. 
Which is a mistake on my part that I haven't written it in the issue. 😞

src/scripts/settings.js
```diff
@@ -32,6 +32,14 @@ document.querySelector('#close-review-prompt').addEventListener('click', dismiss
   link.addEventListener('click', changeActiveTab);
 });
 
+const isNotTogglApp = (val) => {
```
kornelijepetak, line 35: If it's not too much of a hassle, can you please name this parameter to more precisely signify what is coming into the function? `val` is too generic... Is it an `url`?, a `domain`? Find something that is semantically correct.

src/scripts/settings.js
```diff
@@ -389,10 +398,23 @@ const Settings = {
...
+                    .replace('*://*.', '')
+                    .replace('*://', '')
+                    .replace('/*', '');
+                }
+
                 // Handle cases where subdomain is used (like web.any.do, we remove web from the beginning)
-                tmpkey = key.split('.');
-                tmpkey.shift();
-                tmpkey = tmpkey.join('.');
+                if (TogglOrigins[key].url.match(/\/\/\*\./)) {
```
kornelijepetak, line 412: If you think you could name the regex, it'd be much appreciated.
It's hard to read out immediatelly what the `/\/\/\*\./` is 😄, beside looking like a Sioux village 😄

src/scripts/settings.js
langri-sha: The canonical way for expressing no matches is `null` ☝️:

```
val.match(/toggl\./) !== null
```

src/scripts/origins.js
dpsxp: >  but did you actually go thru all of them and checked whether their respective apps are actually under such subdomains?

The problem wasn't this. The button was getting the permission to run on these domains, the only issue is that the integrations was wrongly showing the checkboxes as unchecked. Note that the permission is set based on the `url` not on this key.

kornelijepetak: But if that is the case, what's then the purpose of this change?

dpsxp: To make button settings recognize this integration based on the key. If you see the file will you see that the key matches the domain of the integration but that doesn't necessarily mean that they weren't working is just that logic behind the checkboxes rely on this `key -> url` match to handle the state.

PR #1804, opened 2020-07-30
fix(common): Hide text in minimal button type

rylek90:
We've recently reworked button to use svg and we apparently forgot to hide `Start Timer` text from it.
In few integrations like Toggl Plan it worked by an accident but it screwed Trello checklist for example.

Check Trello checklists.

Before:
image

After:
image

Now matches our [style guide](https://github.com/toggl/toggl-button/blob/master/.github/CONTRIBUTING.md#toggl-button-style-guide)

https://toggl.slack.com/archives/C9LMUPD9T/p1596045257219300

### DIFF
--- a/src/scripts/common.js
+++ b/src/scripts/common.js
@@ -509,7 +509,10 @@ window.togglbutton = {
     }
 
     link.innerHTML = togglButtonSVG;
-    setLinkText(link, 'Start Timer');
+
+    if (params.buttonType !== 'minimal') {
+      setLinkText(link, 'Start Timer');
+    }
 
     link.addEventListener('click', function (e) {
       let opts;


src/scripts/common.js
```diff
@@ -509,7 +509,10 @@ window.togglbutton = {
     }
 
     link.innerHTML = togglButtonSVG;
-    setLinkText(link, 'Start Timer');
+
+    if (params.buttonType !== 'minimal') {
```
dpsxp, line 513: There's an if for the same check a few lines above, maybe we can use it instead of adding a new one 🤔

dpsxp: Oh no nevermind it's `!==` 🤦‍♂️

PR #1815, opened 2020-08-15
Add declarative integrations support

### LINKED ISSUE #1141: Add declarative integrations support
## 🤔 Why?
1. Declarative integrations would mean a lot less code and if we want to keep up with the ever-changing websites we need to decrease the amount of code we have to maintain
2. It'd pave the way for something better: a point-and-click way of adding custom integrations

## 💁 How?
- Use either CSS selectors or XPath to describe the location of elements containing the information for the time entry description and project
- Use template expressions to stitch the information from different elements into description and project
- To describe the button location there'd be an expression pointing to an "anchor" element + extra setting saying how it should be inserted in relation to the anchor (before, after, inside, etc)

## 😱 Challenges
- I'm pretty sure this approach will not work for all integrations so we'd have to keep the old way of doing integrations around
- Perhaps we'd need a mix of declarative and imperative for some integrations, I'm definitely not sure how to do that (`eval`? more like `evil`! _ba dum tss_ 🥁)
- Not sure how the "observe" feature would work here, maybe we'd need to also define the location for an element that'd trigger a refresh

dooart:

🚀  Adds declarative integrations support
:octocat: Refactors the Github integration to use the declarative style.
👼 Refactors the Toggl Plan integration to use the declarative style.
🌈  As a bonus, it'll also fix the automatic project selection in Github.
⚠️ There's no support for auto-filling tags yet, just description and project!

NB: I'm quite aware this will not work for all integrations, but it should work for most of them. If you're still skeptical I can include a couple more tools in this PR before it goes in. ☀️ 

:octocat: Make sure the Github integration still works properly:
- Open any standalone issue. The button should appear above the assignee.
- Open a project board and click an issue. The button should appear inside the issue details view at the right of the screen.
- Start an entry in each view and make sure the description and the project are filled correctly.

👼 Make sure the Toggl Plan integration still works properly:
- Open any task. The button should appear at the top of the modal.
- Start an entry and make sure the description and the project are filled correctly.

Closes #1141

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6206,7 +6206,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -6227,12 +6228,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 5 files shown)


dooart:
closing this pr because it is beyond stale at this point and also inherently incompatible with the new direction for toggl button. i still believe we should explore this in the future, particularly because it'd enable users to write integrations in a much simpler fashion

Issue #1768, opened 2020-06-23
Time entry stopping automicatically and rounding end time up to match pomodoro minutes


dianetoggl:
Windows 10
Chrome 83.0.4103.106
1.62.1

### 🐛 Describe the bug
User has pomodoro set to 60 minutes. Sometimes when they start a time entry (for example at 5.30pm), they will check again at 5.45pm and the entry will have automatically stopped and the end time will be 6.30pm to match the 60 minute pomodoro timer.

### Expected behaviour

Timer wouldn't automatically stop and if it did it wouldn't change the end time to a future time to match pomodoro.

### Other details or context

Unable to reproduce on my end.
[Intercom Link](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/3540765/conversations/27204334049)

dianetoggl:
User uses Toggl Button maily, Mobile app and VERY OCCASIONALLY the webapp

dianetoggl:
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/unassigned/conversations/106481300034281) User has 25 minute pomodoro set - timer stops within a few minutes and rounds up to 25 minutes.

tcrammond:
ℹ️ I think part of the root cause is in here https://github.com/toggl/toggl-button/blob/master/src/scripts/background.js#L866, where "stop the time entry" will trigger, and the duration is always set to exactly the expected Pomodoro duration

This may be fixed in the branch that adds more Pomodoro features https://github.com/toggl/toggl-button/tree/1569/full-pomodoro, but that branch might need a fair bit of finishing off. So maybe an independent fix can be investigated now we know the cause

dpsxp:
It seems that the real problem here isn't the Pomodoro but the time entry stopping before the actual expected time. So even if we fix the Pomodoro duration, the problem will still continue to happen.

dpsxp:
@dianetoggl can you please check with the user if this still happening in the latest release? If so feel free to reopen this issue.

dianetoggl:
Hey @dpsxp - we have another case of this. Finding out version number now
[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/unassigned/conversations/106481304722259)

edtamayo:
One of our users has shared the following, sharing for insight: 


[...]

PR #1825, opened 2020-08-28
fix(pomodoro): Improve pomdoro timer sync on new TEs

### LINKED ISSUE #1801: Pomodoro issue when a new timer is started via the API
* **OS version**: Linux x86_64
* **Browser version**: Chrome 83.0.4103.116
* **Extension version**: 1.64.3

### 🐛 Describe the bug

1) I start p. timer A
2) 10 min remaining on A
3) I call timer B from api
4) chrome extension switches to B and starts a new pomodoro timer with full 25 min left
5) after a few seconds B terminates and pretends it has finished 25 min into the future
### Expected behaviour

Starting a new TE should just reset the Pomodoro interval and not stop the new TE after a few seconds adding the Pomodoro interval time to it's end time.

### Other details or context

Step one, the user referees to the initial timer as the "Pomodoro" timer because they most likely have the countdown view enabled so I guess they refer to it as starting a Pomodoro timer

One nuance here is the fact that they are using an app on their phone to start time entries via the API. They basically have a shortcut for a call to start a running timer but they also use the Toggl Button extension to see Pomodoro.

This issue is possibly related to https://github.com/toggl/toggl-button/issues/1768 and/or https://github.com/toggl/toggl-button/issues/1768#issuecomment-663013721

Intercom thread /106481300034756

dpsxp:
Improve Pomodoro timer sync when creating TEs on other apps/API. The Pomodoro timer should restart when getting new TEs from other services/API

- Turn on Pomodoro 
- I would suggest setting a small interval so you can see the animation better
- Start a new using the extension
- After a few minutes (before the Pomodoro ends) make an API call to start a new running TE.
- The extension should change the running TE to newly created one and set the Pomodoro state based on the current duration of the new TE. 
- If the new TE has a duration already bigger than the Pomodoro interval than the animation should go back to 0% and no alarm should trigger 

#### Sanity check
- Check if Pomodoro still working as expected when only using the extension 
- Updating the duration of TEs should also remove the timer if the new duration is bigger than the interval settings

close #1801 related #1768

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -397,12 +397,13 @@ window.TogglButton = {
       TogglButton.localEntry.id === entry.id;
 
     TogglButton.$curEntry = entry;
+
     if (entry) {
-      if (update) {
-        TogglButton.checkPomodoroAlarm(entry);
-        clearTimeout(TogglButton.$nannyTimer);
-        TogglButton.$nannyTimer = null;
-      } else {
+      TogglButton.checkPomodoroAlarm(entry);
+      clearTimeout(TogglButton.$nannyTimer);
+      TogglButton.$nannyTimer = null;
+
+      if (!update) {
         browser.browserAction.setIcon({
           path: { '19': 'images/icon-19.png', '38': 'images/icon-38.png' }
         });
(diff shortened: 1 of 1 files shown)


kornelijepetak:
@dpsxp sorry for taking this long.

I've noticed one thing:
When I run another TE, the TB popup changes to the newly running TE, however, not in the Pomodoro focus mode. It shows as a regular TE (no circle, just timer log).

Is that intentional?

rylek90 (changes requested):
Can we please rebase it to work with post-rebrand changes? :pray: 

Mostly looks good, I wasn't able to enable focus mode on this branch, though :cry:

dpsxp:
Focus mode should be back now 👍

rylek90 (approved):
LGTM 🚀

src/scripts/background.js
```diff
@@ -688,43 +689,58 @@ window.TogglButton = {
...
-    TogglButton.pomodoroFocusMode = await db.get('pomodoroFocusMode');
-    const interval = parseInt(TogglButton.pomodoroInterval, 10) * 60000;
-    if (duration < interval) {
-      TogglButton.triggerPomodoroAlarm(interval - duration);
-    }
+    TogglButton.triggerPomodoroAlarm(duration);
   },
 
-  triggerPomodoroAlarm: async function (value) {
+  triggerPomodoroAlarm: async function (duration) {
```
kornelijepetak, line 697: Kudos for naming things.

kornelijepetak (approved):
Tried. Works. Shipit.

PR #1841, opened 2020-09-11
Update Icon

Phenome:
Updates Toggl Track icon.


dpsxp (commented):
I don't think we can remove the files. We use it on extension manifest. https://github.com/toggl/toggl-button/search?q=icon-128.png&unscoped_q=icon-128.png

Phenome:
Every icon file was just replaced. They're still there. 🤔

dpsxp:
> Every icon file was just replaced. They're still there. 🤔

🤦 

Just checked the icon, there's a white border now that will mess with the Pomodoro timer progress, so maybe we'll have to change the Pomodoro animation to another color

PR #1850, opened 2020-09-25
fix(manifest): Update the extensions manifests to adhere to the rebranding URLs

### LINKED ISSUE #1784: Toggl Plan is missing from integrations page but the integration is on by default
* **OS version**: 10.15.4
* **Browser version**: 83, Chrome 
* **Extension version**: 1.64.1

Toggl Plan
### 🐛 Describe the bug

The integration is missing from integrations list, but it's on by default. 

### Expected behaviour

User should be able to enable/disable it, even if we decide to have it on by default. 
### Steps to reproduce

1. Go to Integrations page under Settings
2. Look for Toggl Plan 

### Other details or context

[Slack](https://toggl.slack.com/archives/C9LMUPD9T/p1594379026035700)

kornelijepetak:

Updated URLs so that it correctly covers `track.toggl.com` after rebranding.

Test the hell out of this because it has a potential to blow up.
I've tried it locally and it all seems to work correctly, but please test it functionally and also let me know if there's anything amis in the code.

This also updates the manifests, so it will have to go thru the reviews on stores, right?

Incidentally also closes #1784 (make sure to [read the comment](https://github.com/toggl/toggl-button/issues/1784#issuecomment-694431320))

### DIFF
--- a/src/chrome-manifest.json
+++ b/src/chrome-manifest.json
@@ -37,8 +37,7 @@
     "open_in_tab": true
   },
   "permissions": [
-    "*://*.toggl.com/*",
-    "*://toggl.com/*",
+    "*://track.toggl.com/*",
     "webRequest",
     "webRequestBlocking",
     "alarms",
@@ -50,7 +49,7 @@
   ],
   "optional_permissions": ["*://*/"],
   "externally_connectable": {
-    "matches": ["*://*.toggl.com/*", "*://toggl.com/*"]
+    "matches": ["*://track.toggl.com/*"]
   },
   "content_security_policy":
     "script-src 'self'; object-src 'self';",
--- a/src/scripts/lib/db.js
+++ b/src/scripts/lib/db.js
@@ -72,8 +72,8 @@ export default class Db {
   }
 
   async getOriginFileName (domain) {
-    if (process.env.DEBUG && domain.endsWith('toggl.space')) {
-      domain = 'toggl.com';
+    if (process.env.DEBUG && domain.endsWith('track.toggl.space')) {
+      domain = 'track.toggl.com';
     }
     let origin = await this.getOrigin(domain);
     const origins = await this.getAllOrigins();
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,7 +1,7 @@
 # Site settings
 title: "Toggl Button Chrome extension"
 #email:
-url: http://toggl.com
+url: http://track.toggl.com
 description: "Adds Toggl one-click time tracking to popular web tools."
 
 # Color settings (hex-codes without the leading hash-tag)
--- a/src/firefox-manifest.json
+++ b/src/firefox-manifest.json
@@ -34,8 +34,7 @@
     "open_in_tab": true
   },
   "permissions": [
-    "*://*.toggl.com/*",
-    "*://toggl.com/*",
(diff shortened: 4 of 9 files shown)


rylek90 (approved):
Couldn't break things. But I feel like this needs another 👀 

LGTM 🚀

kornelijepetak:
Yes, I definitely want another review. If not several. This is a slippery slope. 😄

src/scripts/lib/db.js
```diff
@@ -72,8 +72,8 @@ export default class Db {
   }
 
   async getOriginFileName (domain) {
-    if (process.env.DEBUG && domain.endsWith('toggl.space')) {
-      domain = 'toggl.com';
+    if (process.env.DEBUG && domain.endsWith('track.toggl.space')) {
+      domain = 'track.toggl.com';
```
shantanuraj, line 76: This should remain `toggl.com` or preferably  we change the `origins.js` file to match the domain.
Currently it's set to `toggl.com` instead of `track.toggl.com`

image

src/chrome-manifest.json
```diff
@@ -37,8 +37,7 @@
     "open_in_tab": true
   },
   "permissions": [
-    "*://*.toggl.com/*",
-    "*://toggl.com/*",
+    "*://track.toggl.com/*",
```
shantanuraj, line 40: I couldn't finish the login flow, I believe we still need access to `toggl.com` sans prefix

Look at these lines
https://github.com/toggl/toggl-button/blob/ce3fc5d34f317e3564ecbd82783bed2296498bab/webpack.config.js#L76
https://github.com/toggl/toggl-button/blob/16155d8560dfa2d9d6b1e550e6cc87fbebf989e8/src/scripts/components/LoginPage.tsx#L198

The extension needs access to `toggl.com/track/toggl-button-login/` to communicate with it and be informed when the login completes
https://github.com/toggl/toggl-button/blob/1f335016dc614a5289eac5b471bd9fbc8c93de2f/src/scripts/content/toggl.js#L37-L48

But this content script won't work unless it has access to the public-web page

shantanuraj (changes requested):
On first run the extension seemed to work fine,

But since granted permissions aren't revoked on loading a new extension build.

I created a new Chrome profile to test the extension and couldn't get the login-flow to work as intended the login still completes due to shared cookies, but the post-login page isn't shown and the user will remain on the public-web page.

See more [here](https://toggl.slack.com/archives/C029W49FT/p1601307988024000)

Also noted an additional code-change regarding the domain naming in origins list

kornelijepetak:
@shantanuraj it seems that the problem you have described happens on `master` as well. Can you please check it?

I was trying to find out which part of my PR breaks things, but as I was removing my changes, nothing different was happening. And then I tried:

* creating another chrome profile
* switch to master
* `yarn start`
* try to log in

The popup is logged in, I can indeed start a TE, but the login page will just remain there, with my email/password still entered, just as you described.

shantanuraj:
@kornelijepetak I tried the steps you described and shared my recording [here ](https://toggl.slack.com/archives/C9LMUPD9T/p1601560331055100?thread_ts=1601558105.054900&cid=C9LMUPD9T)

I am not able to reproduce this behaviour on `master`

PR #1855, opened 2020-10-06
fix(timer): Stop current time entry before starting

### LINKED ISSUE #1846: Issue with some integrations when required fields for TEs are enabled
* **OS version**: OS X 10.14.6
* **Browser version**: Chrome 85.0.4183.102
* **Extension version**: 1.66.3

https://app.asana.com
https://trello.com/

### 🐛 Describe the bug

A user reached out saying that (when they have required fields turned on) if they have a time running already via the Track extension and they click on the Track button to start a new time entry from Asana or Trello they previous timer doesn't get saved but it actually gets renamed and both timers are basically turned into one when the currently running one is stopped. 

We managed to get to the point where asking them to turn of required fields for time entries and they confirmed that this helped. But they also confirmed (and shared a video of this issue happening coroborating their actions) that they are using all the required fields when starting a timer via the Track extension. 

They also say that this is not an issue with other tools they use the Track button with. 

You can check  out the video **[here](https://drive.google.com/file/d/13j1ZlIUd6-98DWZR4Ql1THYfQ9K47CLn/view?usp=sharing)**. 

### Expected behaviour

When they start a new running timer using the Track button from the Asana or Trello web app it should save the old one. In the event that required fields are not filled out they should get a warning. 

### Steps to reproduce

I'm afraid that I was not able to reproduce this issue with required fields turnd on in either one of those apps.

dpsxp:
Stop current time entry before starting a new one to avoid issues with time entry constraints

- Set some constraints to create time entries first
- Now go and create a new Time Entry using the button
- Try to start a new one without stopping the current one
- The first TE should stop and a new one should start 👍 

close #1846

### DIFF
--- a/src/scripts/common.js
+++ b/src/scripts/common.js
@@ -543,8 +543,11 @@ window.togglbutton = {
         };
       }
       togglbutton.element = e.currentTarget;
+
       browser.runtime
-        .sendMessage(opts)
+        // Stop current time entry before starting a new one
+        .sendMessage({ type: 'stop' })
+        .then(() => browser.runtime.sendMessage(opts))
         .then(togglbutton.addEditForm);
 
       return false;


kornelijepetak (approved):
LGTM

rylek90 (approved):
GTML 🚀

PR #1863, opened 2020-10-16
fix(styling): Prevent gray button

### LINKED ISSUE #1834: Todoist integration not working
* **OS version**: 10.15.6
* **Browser version**: Chome 85
* **Extension version**: 1.66.1

Todoist 
### 🐛 Describe the bug

Here's how the button looks 
image

### Expected behaviour
The button should show correctly. 
### Other details or context

[Slack](https://toggl.slack.com/archives/C9LMUPD9T/p1599556983200900)

### LINKED ISSUE #1835: Button is completely grey in Asana integration
* **OS version**: 10.15.6
* **Browser version**: Chrome 85
* **Extension version**: 1.66.1

Asana (same behavior in Todoist)

### 🐛 Describe the bug

The button is entirely grey until you click on it and it turns purple

Screenshot 2020-09-08 at 11 57 55

### Expected behaviour

The button show correctly with our logo

### Other details or context

[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2759100/conversations/106481303544488)

### LINKED ISSUE #1862: Toggl extension icon appearing as a grey dot in tools
* **OSX 10.5.7**: 
* **Chrome 86.0.4240.75**: 
* **1.66.7**: 

Asana https://app.asana.com) but affecting other services. Shantanu has said the DOM integration also has problems, see [Slack discussion
](https://toggl.slack.com/archives/C9LMUPD9T/p1602649828146500)

### 🐛 Describe the bug

When using the extension with Asana, a grey dot is appearing with no information to suggest it is Toggl, confusing users about how to start a timer.

### Expected behaviour

An obvious Toggl icon should be displayed, for intuitive use.

### Steps to reproduce

I haven't reproduced but a video is [included here](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2097915/conversations/106481304714330) and see the [Slack discussion](https://toggl.slack.com/archives/C9LMUPD9T/p1602649828146500) for another example

kornelijepetak:

Fixes the way `path` SVG element is selected, consequently fixing the gray button issue.

The issue was the fact that the selector `... svg path` would apply to BOTH `<path>` elements of the svg: one is the background circle, the other one is the power icon.

The solution I provided fixes the problem, but if we ever change the svg, we could create another problem. But I saw no other way to meaningfully identify particular `<path>` element inside an svg. Alternatively, we could add a `class` attribute to the `<path>` elements inside the SVG file itself and then select them by class. But that would require new designs to be aware of this procedure and provide the same with new files.

If you think this (☝️) way is better, let me know. I have a commit prepared already. Just didn't know if I should take this approach so I didn't push it.

Make sure you can see pink/black button when a TE is running and gray/white button when it's not. Preferrably not only in Todoist, but also in some other integrations as well (Asana is another known failing integration).

Closes #1834 
Closes #1835
Closes #1862


kornelijepetak:
I have pushed the second approach as well.
When we decide on the approach I will either:

* squash the commits before rebasing
* reset the branch to contain only 1st commit (+3rd which is just a typo fix)

Phenome:
I agree that `:first-child` and `:last-child` selectors are not really stable. But I still think it's OK to go as is to fix this issue.
My take is we'll have to take this into consideration for future updates to the SVG. Both if we go with the class-based proposed approach, or with `first` and `last` selectors here. So might as well go as is.

Phenome (approved):
LGTM 🚀

PR #1872, opened 2020-11-01
feat(gitea): Implement Gitea integration

belidzs:

Adds a new integration with Gitea. Gitea is a community managed lightweight code hosting solution written in Go. It is published under the MIT license. 

Go to https://gitea.com or https://try.gitea.io to see it in action.

image

### DIFF
--- a/src/scripts/content/gitea.js
+++ b/src/scripts/content/gitea.js
@@ -7,6 +7,7 @@ togglbutton.render('.time-desc:not(.toggl)',
     const link = togglbutton.createTimerLink({
       className: 'gitea',
       description: descriptionSelector,
+      tags: tagsSelector,
       projectName: projectSelector,
     });
     $container.appendChild(link);
--- a/src/scripts/content/gitea.js
+++ b/src/scripts/content/gitea.js
@@ -1,0 +1,36 @@
+'use strict';
+/* global togglbutton, $ */
+
+togglbutton.render('.time-desc:not(.toggl)',
+  { observe: true },
+  $container => {
+    const link = togglbutton.createTimerLink({
+      className: 'gitea',
+      description: descriptionSelector,
+      projectName: projectSelector,
+    });
+    $container.appendChild(link);
+  }
+);
+
+function descriptionSelector () {
+  const description = document.getElementById('issue-title');
+  const issueId = description.previousElementSibling;
+  return `${issueId.textContent} ${description.textContent}`;
+}
+
+function projectSelector () {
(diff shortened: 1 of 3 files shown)


src/scripts/content/gitea.js
```diff
@@ -0,0 +1,36 @@
...
+  const issueId = description.previousElementSibling;
+  return `${issueId.textContent} ${description.textContent}`;
+}
+
+function projectSelector () {
+  const $project = document.getElementsByClassName('repo-title')[0].children[3];
+  return $project ? $project.textContent.trim() : '';
+}
+
+function tagsSelector () {
```
dpsxp, line 27: It looks like you forgot to add this function to one of the params when creating the button link.

dpsxp:
Thanks for taking the time to add more integrations to this project. I left a comment about missing tags, it seems that you have the code ready for it but you forgot to add it as one of the parameters.

For future reviewers here's a quick link that can help to see the new integration in action https://try.gitea.io/AyudiaHumera/AgenBola/issues/1

belidzs:
@dpsxp thanks for attending to this PR so quickly.

I have actually developed and tested the tag-parsing capability as you have noticed, but I've found that no matter how I pass the extracted tags to the button, it never shows up in the selector.

I suspected that the tags maybe need to exist in advance but since I couldn't test it I just didn't wire it in. If you think that this should work as is (passing the tags as an array of string) I can push another commit in no time.

src/scripts/content/gitea.js
belidzs: I have actually developed and tested the tag-parsing capability as you have noticed, but I've found that no matter how I pass the extracted tags to the button, it never shows up in the selector.

I suspected that the tags maybe need to exist in advance but since I couldn't test it I just didn't wire it in. If you think that this should work as is (passing the tags as an array of string) I can push another commit in no time.

dpsxp: Indeed the code doesn't seem to support tags when creating entries using the `createTimerLink` function. But it does sound like a bug. I think you can pass the `tagSelector` as one of the params since you already have it done, and when the code gets fixed you'll have the tags working as a side effect.

dpsxp (approved):

belidzs:
I've passed the `tagsSelector` function to `togglbutton.createTimerLink()` as per your suggestions.

I think this PR is ready to merge at this point.

dpsxp:
Thanks for your work on this @belidzs , we really appreciate 🤗

PR #1861, opened 2020-10-13
fix(background): Fix adding custom integrations URLs

rylek90:
Please remember the [Contributing Guidelines](https://github.com/toggl/toggl-button/blob/master/.github/CONTRIBUTING.md) :heart:

## :star2: What does this PR do?

Fixes custom URLs.

## :bug: Recommendations for testing
On master:
* install extension
* go to settings
* do not enable any integrations from list
* set custom URL to `facebook.com`, pick any integration you want
* go to facebook.com, see content script for chosen integration not loaded
* uninstall extension
* switch to this branch
* repeat setting custom URL
* go to facebook.com
* your script should now be loaded

PR #1874, opened 2020-11-02
Switch to wildcards instead of www

kornelijepetak:

#1861 broke integrations that are starting with `www.` in origins.

This PR switches them to the wildcards and explicitly sets the `file` so that the scripts can be loaded properly on respective websites.

How about you be a good boy and test all 4 integrations? 😆 

Fixes what was broken by #1861


rylek90 (approved):
Scripts are loaded :ok_hand:

PR #1881, opened 2020-11-11
chore(toggl-plan): Use additional data-attr selector to inject

rylek90:
Plan has now their own integration with Track, so we must disable button if it's enabled on their side.

Just verify if it still injects properly on Plan. You can also create two elements (one with `data-track-inject-button` attr) and verify selector.

See: https://toggl.slack.com/archives/C0198AEAZ7T/p1605003706176300


kornelijepetak (approved):
oh, if only all PRs were this easy to review 😄

dpsxp (approved):

PR #1907, opened 2021-01-19
More advanced Asana sub-task support

willrowe:

This adds more advanced support for Asana sub-tasks. An Asana sub-task is a task that is nested under another task. Currently top-level tasks are supported, but they are slightly different than sub-tasks. Sub-tasks display the projects that their top-level ancestor belongs to in a different location, so this PR allows this location to be taken into consideration.

This PR also includes the name of ancestor tasks in the description in the same way it is displayed by Asana.

This does not change any functionality for top-level tasks, missing ancestor elements do not match anything and are ignored.

### DIFF
--- a/src/scripts/content/asana.js
+++ b/src/scripts/content/asana.js
@@ -138,7 +138,10 @@ togglbutton.render(
     }
 
     const descriptionSelector = () => {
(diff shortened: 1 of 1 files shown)


willrowe:
Is there any interest in adding these features?

bytrangle:
> Is there any interest in adding these features?

I appreciate your effort, but I don't think this project has been active recently.

PR #1924, opened 2021-03-04
fix(atlassian jira): Select the correct Jira issue key in board sidebar

### LINKED ISSUE #1818: Projects in Jira no longer being autopopulated
* **OS version**: MacOS 10.15.6
* **Browser version**: 84.0.4147.125
* **Extension version**: 1.64.7

### 🐛 Describe the bug
Toggl button shows on Jira but projects must be manually selected.

### Expected behaviour
Project should be automatically populated

### Steps to reproduce

Unfortunately, don't have access to Jira so can't reproduce.

### Other details or context

[case](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/2144650/conversations/106481300011965)

### LINKED ISSUE #1912: Description pre-filled with Jira epic key instead of issue key
* **OS version**: Ubuntu 20.04
* **Browser version**: Chrome Version 87.0.4280.88 (Official Build) (64-bit)
* **Extension version**: 1.69.0

Jira Cloud - https://atlassian.net

### 🐛 Describe the bug

Atlassian have recently rolled out a new interface for issues in Jira. Since that update, starting the Toggl time tracking from an issue will prefill the time entry description using that issue's epic key, instead of the issue key. This seemed to work correctly prior to that update. This problem only occurs with the side view of an issue when on the backlog, not when opening the full page view of an issue or the popup view on the active sprint/kanban.

So for an issue with key ABC-2, name "My task" and linked to epic ABC-1. The description in the Toggl button when started on the issue will prefill to: "ABC-1 My task", rather than "ABC-2 My task".

### Steps to reproduce

1. Start a time entry on any issue assigned to an epic when on the backlog. In this example, the issue key is TRAIN-619 and the epic it is attached to is TRAIN-221.

Screenshot from 2021-02-02 12-38-48

2. The Toggl dialog that drops down will have the description pre-filled using the wrong key. Here you can see the epic key of TRAIN-221 has been added to the issue name as the description.

Screenshot from 2021-02-02 12-35-57

7ochem:

In the Jira Board issue sidebar, when an issue has a parent issue/epic, then when The Toggle button gets placed, it takes the parent issue number (being the first `<a />`) instead of the correct issue number.

Screenshot from 2021-03-04 13-22-22

The issue number in the description will be PROJ-1, which is wrong. This PR changes the code so it will pick the last one, so PROJ-3 here.

I have tested this on a SaaS Jira instance (at https://*.atlassian.net/). I've tried to test this on public Jira boards but couldn't find any where the piece of code would be applicable to in the first place.

Fixes #1912

### DIFF
--- a/src/scripts/content/atlassian.js
+++ b/src/scripts/content/atlassian.js
@@ -12,7 +12,7 @@ togglbutton.render(
     const rootElem = $('#ghx-detail-view');
     const container = createTag('div', 'jira-ghx-toggl-button');
     const titleElem = $('[spacing] h1', rootElem);
-    const numElem = $('[spacing] a', rootElem);
+    const numElem = $('[spacing]:eq(0) a:last', rootElem);
     const projectElem = $('.bgdPDV');
     let description = titleElem.textContent;
     if (numElem !== null) {


7ochem:
@shantanuraj is there anything you need to get this PR moving?

PR #1936, opened 2021-04-12
Bugsnag spring cleaning

### LINKED ISSUE #1938: Bugsnag spring cleaning
This is a placeholder issue to link to bugsnag reports.

403 error on getting user data (error `5f0ea96fe8175c001787b602`)
Checks for element on addEventListener error (error `5dde294cb118e1001ae5a1d7`)
Promised response from onMessage listener went out of scope (error `6045decb96760f0018a4d639`)

Phenome:
Treats most common Bugsnag errors.

* Upgrades Bugsnag library to `7.9.2`. Check https://github.com/bugsnag/bugsnag-js/blob/master/UPGRADING.md#6x-to-7x
* Adjusts path separators so errors are merged across OSs
* Omits common 403 error on getting user data (error `5f0ea96fe8175c001787b602`)
* Checks for element on addEventListener error (error `5dde294cb118e1001ae5a1d7`)
* Adds more debug info to URL error (error `6045ddaba7d01e001798e085`)
* Fixes Promised response from onMessage listener went out of scope (error `6045decb96760f0018a4d639`)
  Helper docs on this one:
  `onMessage `[should](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#sending_an_asynchronous_response_using_a_promise) either return `true`, and then use `sendResponse`, OR it should return a promise, and then use `resolve`. We were returning a promise and also trying to use `sendResponse` so I refactored it. Hopefully this solves it.

* Check each commit separately 🙏🏻
* Make sure you give https://github.com/bugsnag/bugsnag-js/blob/master/UPGRADING.md#6x-to-7x a read against commit ba75cda

----

Closes #1938

### DIFF
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -246,11 +246,11 @@ window.TogglButton = {
               TogglButton.setCanSeeBillable();
               ga.reportOs();
             } else {
-              bugsnagClient.notify(new Error(`Fetch user failed ${xhr.status}`), {
-                metaData: {
+              bugsnagClient.notify(new Error(`Fetch user failed ${xhr.status}`), evt => {
+                evt.addMetadata('general', {
                   status: xhr.status,
                   responseText: xhr.responseText
-                }
+                });
               });
               TogglButton.setBrowserActionBadge();
               resolve({
@@ -283,9 +283,9 @@ window.TogglButton = {
 
   updateBugsnag: function () {
     // Set user data
-    bugsnagClient.user = {
+    bugsnagClient.setUser({
       id: TogglButton.$user.id
-    };
+    });
   },
 
   setCanSeeBillable: function () {
@@ -316,7 +316,7 @@ window.TogglButton = {
     try {
       TogglButton.websocket.socket = new WebSocket('wss://stream.toggl.com/ws');
     } catch (error) {
-      bugsnagClient.notify(error, { context: 'websocket' });
+      bugsnagClient.notify(error, evt => { evt.context = 'websocket'; });
       TogglButton.retryWebsocketConnection();
       return;
     }
@@ -336,7 +336,7 @@ window.TogglButton = {
         return TogglButton.websocket.socket.send(data);
       } catch (error) {
         if (process.env.DEBUG) console.log(error);
-        bugsnagClient.notify(error, { context: 'websocket' });
+        bugsnagClient.notify(error, evt => { evt.context = 'websocket'; });
       }
     };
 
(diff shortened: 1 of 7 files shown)


src/scripts/background.js
```diff
@@ -246,11 +246,11 @@ window.TogglButton = {
               TogglButton.setCanSeeBillable();
               ga.reportOs();
             } else {
-              bugsnagClient.notify(new Error(`Fetch user failed ${xhr.status}`), {
-                metaData: {
+              bugsnagClient.notify(new Error(`Fetch user failed ${xhr.status}`), evt => {
```
rylek90: According to `UPGRADING.md` we can just use static here `import Bugsnag from '@bugsnag/js'` but I don't really like importing some magic singleton :shrug: 
So if it still works then I'm happy with it.

src/scripts/lib/bugsnag.ts
```diff
@@ -16,6 +16,7 @@ function processError(err: Error) {
       /.*(moz-extension|chrome_extension|chrome-extension|file):\/\/.*\/scripts\/(.*)/gi,
       'togglbutton://scripts/$2'
     );
+    frame.file = frame.file.replace(/\\/g, '/');
```
rylek90, line 19: Today we know what it is about. Tomorrow we may not. Can we leave some comment? :pray:

src/scripts/popup.js
```diff
@@ -566,22 +566,25 @@ document.addEventListener('DOMContentLoaded', function () {
...
-      .querySelector('#toggl-button-duration')
-      .addEventListener('blur', function (event) {
-        const value = event.target.value || '';
-        const parsedInput = parseDuration(value).asSeconds();
-        event.target.value = secToHhmmImproved(parsedInput, { html: false });
-      });
+    // TODO: Properly fix this
+    // Bugsnag error 5dde294cb118e1001ae5a1d7
+    const durationElement = document.querySelector('#toggl-button-duration');
+    if (durationElement) {
```
rylek90, line 572: :see_no_evil: solid, LGTM :rocket:

rylek90 (approved):
Seems to be still working, changes looks solid. Let's give it a spin :rocket:

src/scripts/background.js
Phenome: The upgrading docs [also state](https://github.com/bugsnag/bugsnag-js/blob/master/UPGRADING.md#new-static-interface) that:
> You can choose to hold on to the Client returned by Bugsnag.start(), or not. After ensuring Bugsnag.start() is called first, you can call any Client method on the static interface, which forwards the method call onto the initialized client

Since we're also returning some other object from our wrapper library if we can't initialize it, I though it best to still use the returned client 👍🏻

PR #1968, opened 2021-05-28
Toggl labs/close on start

almirbi:

* block tabs on start time entry
* add block page
* add timer to block page

Go to Settings -> Distraction blocker, make sure to add a URL, like https://www.youtube.com/
Open that URL, like https://www.youtube.com/, in a new tab
Start time entry for anywhere in the extension

Should show block page, still not synced though


src/scripts/website-blocking/WebsiteBlocking.ts
```diff
@@ -3,7 +3,11 @@ import browser from 'webextension-polyfill';
 import WebsiteBlockingApi from "./WebsiteBlockingApi";
 import WebsiteBlockingConverter from "./WebsiteBlockingConverter";
 
-const db = browser.extension.getBackgroundPage().db;
+const db = () => browser.extension.getBackgroundPage().db;
```
EternallLight, line 6: what's the difference? it isn't fully initialised here?

src/scripts/website-blocking/WebsiteBlocking.ts
```diff
@@ -3,7 +3,11 @@ import browser from 'webextension-polyfill';
...
+
+function youShallNotPass(tabId) {
```
EternallLight, line 8: let's move this to yet another static function
this class is going to be shrunk down anyway

src/scripts/website-blocking/WebsiteBlocking.ts
```diff
@@ -117,8 +121,33 @@ class WebsiteBlocking {
...
-}
 
-export default WebsiteBlocking;
+  static async closeOnStart() {
+    
+    const websiteBlockingList = JSON.parse(await db().get('websiteBlockingList') || '[]');
 
+    const blockedSites = websiteBlockingList.map(entry => entry.url);
 
+    console.log(blockedSites)
```
EternallLight, line 131: I forget to remove those often, too 😀

src/scripts/website-blocking/WebsiteBlocking.ts
```diff
@@ -117,8 +121,33 @@ class WebsiteBlocking {
...
+    if (blockedSites.length > 0) {
+      const blockedTabs = await browser.tabs.query({ url: blockedSites });
+      blockedTabs.forEach(tab => youShallNotPass(tab.id));
+    }
+  };
+
+  static async blockSite(tabId, changeInfo) {
+    const url = changeInfo.pendingUrl || changeInfo.url;
+
```
EternallLight, line 140: Let's put early return here if `url` is falsy.
This function seems to be triggered a lot - e.g. when title is changed on a tab.

EternallLight (changes requested):

EternallLight:
And stop button doesn't seem to work for me
image
image

src/scripts/website-blocking/WebsiteBlocking.ts
almirbi: I think it's too early when I use it in backgorund.js, it was failing.

EternallLight (approved):

PR #1970, opened 2021-05-31
Fix blocking full URLs 

EternallLight:

The current implementation does block `https://www.instagram.com/`, but it fails to block `https://www.instagram.com/accounts/login/`.
The same is for `https://www.youtube.com` and direct links to videos like `https://www.youtube.com/watch?v=DYzXKZn_Z3w`.

Try opening `https://www.youtube.com/watch?v=DYzXKZn_Z3w` while blocking YT and tracking time.


almirbi (approved):
LGTM 🚀

PR #1974, opened 2021-06-23
fix(any.do): Any.do integration

tomcat0090:
- Update selectors

But only task list.
Not showed in Task card.

＃1799


tomcat0090:
Recreate pull request.
This one have mistake.

PR #1975, opened 2021-06-23
fix(any.do): Any.do integration

tomcat0090:
＃1799

But only task list.
Not showed in Task card.


tomcat0090:
I created new pull request

PR #1984, opened 2021-07-16
Notion fixes

### LINKED ISSUE #1983: Notion button state doesn't get updated
* **OS version**: All
* **Browser version**: All
* **Extension version**:  1.70.2

### 🐛 Describe the bug

An extension question, a report that Notion doesn't update the button (in-page) state while running.

### Expected behaviour

Button should update to "Stop timer" while, yeah, timer is running

### Steps to reproduce

1. Go to Notion and attempt to start a timer. T

### Other details or context

[Slack](https://toggl.slack.com/archives/C015ZEJ3YKF/p1626207652125500)

rylek90:

Two things, actually. 

First commit is simple enough. It just fixes Notion new selectors, so that description is added properly.

The second commit fixes the way we indicate if button is running or not. Because we do some crazy mumbo-jumbo here:
https://github.com/toggl/toggl-button/blob/d1491c27442aa5fe5214bd39d29464b874b1f110/src/scripts/common.js#L637-L647

In order to determine which button instance/link is currently running. There was a problem if description was obtained *after* button creation (with a function, in dynamic page load). So it resets link `title` attribute after it creates time entry. 

Just check out if both issues are fixed + do some sanity test on other integrations. 

Closes ##1982
Closes #1983

### DIFF
--- a/src/scripts/common.js
+++ b/src/scripts/common.js
@@ -583,7 +583,6 @@ window.togglbutton = {
           service: togglbutton.serviceName
         };
       } else {
-        togglbutton.activateTimerLink(link);
         opts = {
           type: 'timeEntry',
           respond: true,
@@ -596,6 +595,8 @@ window.togglbutton = {
           url: window.location.href,
           container: params.container || ''
         };
+        link.title = opts.description + (opts.projectName ? ' - ' + opts.projectName : '');
+        togglbutton.activateTimerLink(link);
       }
       togglbutton.element = e.currentTarget;
 
--- a/src/scripts/content/notion.js
+++ b/src/scripts/content/notion.js
@@ -1,13 +1,13 @@
 'use strict';
 
-// Selectors here are madness, it works for as of Dec 4th 2019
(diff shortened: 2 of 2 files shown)


nunofmn (approved):
LGTM! 👍 🚀

shantanuraj:
Unrelated to the PR

> https://github.com/toggl/toggl-button/blob/d1491c27442aa5fe5214bd39d29464b874b1f110/src/scripts/common.js#L637-L647

but there's a bug in the sorter, both if conditions are identical

```tsx
  if (a.title.length > b.title.length) return -1; 
  if (b.title.length < a.title.length) return 1; 
```

I think the intent was

```diff
  if (a.title.length > b.title.length) return -1; 
- if (b.title.length < a.title.length) return 1; 
+ if (a.title.length < b.title.length) return 1; 
```

but I think a simpler solution would be to have it be.

```tsx
  return b.title.length - a.title.length;
```

oliviale:
lol who's this guy

Issue #1986, opened 2021-07-20
Toggl Button login not working if the WS has too many projects or clients


alexcr:
* **OS version**: Windows 10
* **Browser version**: Chrome 91.0.4472.124
* **Extension version**: 1.70.2

### 🐛 Describe the bug

Login fails due to `QuotaExceededError` on these lines:
https://github.com/toggl/toggl-button/blob/1e78c202c303dce18f816e9a070333fb06e0f28e/src/scripts/background.js#L220-L221

### Expected behaviour

Login should work!

TBD how exactly we'll handle this.

Note that the stored data is used here:
https://github.com/toggl/toggl-button/blob/9ae9a81d92aff4d12c97d6547e2ff8e784c0a720/src/scripts/settings.js#L214-L219

### Steps to reproduce

Steps to reproduce the behaviour:
1. Simply try to login to the extension on a WS that has >5MB data of projects/clients (e.g. user 6671101)

### Other details or context

Fore more info see Slack convo:
https://toggl.slack.com/archives/C9LMUPD9T/p1623793352149700

Phenome:
I think we should switch to IndexedDB for storage. Much more quota there.

PR #1989, opened 2021-07-21
feat(db) Store projects, clients and other clientside data in IndexedDB to avoid LocalStorage limitation

EternallLight:

For cases where serialised data of projects and clients exceeds 5 MB, the extension wouldn't log in.

This PR changes storing of projects and clients to IndexedDB thereby avoiding the given limitation and not introducing new required permissions like other storages do (`unlimitedStorage`).

Since IndexedDB API is verbose and we might make heavy use of it in the future, I took the liberty of introducing the `localforage` library to the extension. 

0) Generate a user with lots of projects and clients
1) Check if the current version of the extension can login. If it can, you probably don't have enough projects/clients.
2) Check if all pages work correctly - extension, settings, etc, and no issues in console. The `TrackExtension` object store (localforage database) should be available in IndexedDB dev tools settings. Check the console for new scary errors. 
3) Check if the database gets cleared out when logging out

#1986

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,6 @@
     "@emotion/styled": "10.0.10",
     "date-fns": "1.30.1",
     "keycode": "2.2.0",
-    "localforage": "1.9.0",
     "lodash.map": "4.6.0",
     "lodash.reduce": "4.6.0",
     "moment": "2.24.0",
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "@emotion/styled": "10.0.10",
     "date-fns": "1.30.1",
     "keycode": "2.2.0",
+    "localforage": "1.9.0",
     "lodash.map": "4.6.0",
     "lodash.reduce": "4.6.0",
     "moment": "2.24.0",
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -221,10 +221,6 @@ window.TogglButton = {
               }
 
               TogglButton.updateTriggers(entry);
-              await Promise.all([
-                db.setLocal('projects', projectMap),
-                db.setLocal('clients', clientMap)
-              ]);
               TogglButton.$user = resp.data;
               TogglButton.$user.time_entries = (TogglButton.$user.time_entries || [])
                 .map((te) => {
@@ -235,6 +231,10 @@ window.TogglButton = {
                     tid: te.tid || null
                   };
                 });
+              await Promise.all([
+                db.setLocal('projects', projectMap),
+                db.setLocal('clients', clientMap)
+              ]);
               TogglButton.$user.projectMap = projectMap;
               TogglButton.$user.clientMap = clientMap;
               TogglButton.$user.clientNameMap = clientNameMap;
@@ -259,8 +259,6 @@ window.TogglButton = {
                     responseText: xhr.responseText
                   });
                 });
-              } else {
-                await db.clearLocal();
               }
               TogglButton.setBrowserActionBadge();
               resolve({
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -259,6 +259,8 @@ window.TogglButton = {
                     responseText: xhr.responseText
                   });
                 });
+              } else {
+                await db.clearLocal();
               }
               TogglButton.setBrowserActionBadge();
               resolve({
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -232,8 +232,8 @@ window.TogglButton = {
                   };
                 });
               await Promise.all([
-                db.setLocal('projects', projectMap),
-                db.setLocal('clients', clientMap)
+                db.setLocal(`${TogglButton.$user.id}-projects`, projectMap),
+                db.setLocal(`${TogglButton.$user.id}-clients`, clientMap)
               ]);
               TogglButton.$user.projectMap = projectMap;
               TogglButton.$user.clientMap = clientMap;
--- a/src/scripts/background.js
+++ b/src/scripts/background.js
@@ -169,7 +169,7 @@ window.TogglButton = {
       TogglButton.ajax('/me?with_related_data=true', {
         token: token,
         baseUrl: TogglButton.$ApiV8Url,
-        onLoad: function (xhr) {
+        onLoad: async function (xhr) {
           let resp;
           const projectMap = {};
           const clientMap = {};
@@ -221,8 +221,10 @@ window.TogglButton = {
               }
 
               TogglButton.updateTriggers(entry);
-              localStorage.setItem('projects', JSON.stringify(projectMap));
-              localStorage.setItem('clients', JSON.stringify(clientMap));
+              await Promise.all([
+                db.setLocal('projects', projectMap),
+                db.setLocal('clients', clientMap)
+              ]);
               TogglButton.$user = resp.data;
               TogglButton.$user.time_entries = (TogglButton.$user.time_entries || [])
                 .map((te) => {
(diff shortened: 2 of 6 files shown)


LulzAugusto:
Hmmm `localForage` is really good but it does a lot also so its size is in the kB order of magnitude. Would consider something even simpler like [idb-keyval](https://github.com/jakearchibald/idb-keyval) instead (~450 bytes)?

alexcr:
>Hmmm localForage is really good but it does a lot also so its size is in the kB order of magnitude. Would you consider something even simpler like idb-keyval instead (~450 bytes)?

@LulzAugusto @EternallLight
I found this minimal IndexedDB example and it really doesn't seem that complicated... Unless it doesn't cover our needs, I would just go without any lib and save those 450 bytes: https://gist.github.com/JamesMessinger/a0d6389a5d0e3a24814b

LulzAugusto:
Yeah or that, as long as we don't need to call `db.transaction` and `tx.objectStore` in multiple places I'm fine. :)

package.json
```diff
@@ -92,6 +92,7 @@
     "@emotion/styled": "10.0.10",
     "date-fns": "1.30.1",
     "keycode": "2.2.0",
+    "localforage": "1.9.0",
```
alexcr, line 95: See other comments about using this lib.

I would suggest creating very simple getter/setter functions based on [this minimal example](https://gist.github.com/JamesMessinger/a0d6389a5d0e3a24814b) - but using Promises so we can call them with `await` instead of passing callbacks around, so the usage would be the same as currently: `await db.getLocal(...)`. Also, I don't think we need the index.

For example, something like this (note that I haven't tested this, nor changed it to const, single quotes, etc):
```javascript
getLocal (key) {
  return new Promise((resolve, reject) => {
    // Start a new transaction
    var db = open.result;
    var tx = db.transaction("MyObjectStore", "readwrite");
    var store = tx.objectStore("MyObjectStore");

    // Query the data
    const query = store.get(key);
    query.onsuccess = function() {
      resolve(query.result);
    };

    // Close the db when the transaction is done
    tx.oncomplete = function() {
      db.close();
    };
  })
}
```

Then something similar for the setter, and this on the initialization:
```javascript
// Open (or create) the database
var open = indexedDB.open("MyDatabase", 1);

// Create the schema
open.onupgradeneeded = function() {
    var db = open.result;
    var store = db.createObjectStore("MyObjectStore", {keyPath: "id"});
};
```

alexcr (changes requested):

src/scripts/lib/db.js
```diff
@@ -259,12 +264,20 @@ export default class Db {
       });
   }
 
-  getLocal (key) {
-    const collection = localStorage.getItem(key);
-    if (!collection) {
-      return null;
+  async getLocal (key) {
+    const localStorageData = localStorage.getItem(key);
```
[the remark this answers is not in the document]
EternallLight: Yeah, it could have returned the stale projects/clients for the first time after update. This method also had another issue with returning unparsed JSON, but it's all fixed up now. Thanks for heads-up! 👍🏻

EternallLight:
Guys, thanks for thorough reviews, I've removed localforage in favour of simplistic helper functions.

One thing I'm not sure that if I should open the database on every function run or keep the database always open in the closure. Considering how little we're using this now, I ended up opening and closing the database connection on each request. Let me know if it's better to be done otherwise.

alexcr (dismissed):
Probably better to revert changes in `package-lock.json` now that you removed the lib?

But apart from that, everything looks good now! There's just a very minor code suggestion.

I've tested the extension on Chrome and Firefox with a big WS and it seems to work fine 💯

src/scripts/lib/dbUtils.ts
```diff
@@ -0,0 +1,63 @@
...
+  const db = open.result;
+  db.createObjectStore("LocalData", {keyPath: "id"});
+};
+
+export const getIDBItem = (key) => {
+  return new Promise((resolve, reject) => {
+    const open = indexedDB.open(DB_NAME, DB_VERSION)
+    open.onsuccess = () => {
+      const db = open.result;
+      const transaction = db.transaction("LocalData", "readwrite");
```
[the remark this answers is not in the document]
EternallLight: Right, good catch

EternallLight:
Reverted `package-lock.json`, changed the IDB flag to `readonly` on get operations.
Today I also noticed a weird thing - my yesterday's token expired, and after logging in with a different user I somehow ended up with projects of the previous user. As a safety net, I added clearing IDB data when received 403.

src/scripts/background.js
```diff
@@ -2003,6 +2008,7 @@ window.TogglButton = {
       let error;
       let errorSource;
       try {
+        console.log(request.type);
```
alexcr, line 2011: I think you forgot to remove this :)

alexcr (commented):
Hmm, your solution for the cross-user storage issue seems to work, but I'm a bit worried that there might be other cases that we're not seeing where this would also happen 🤔 Maybe other error codes, or clearing the session manually, something like that... I couldn't find another way to hit the issue, it's just a random thought.

Maybe it would be better to store the user ID along with the data, and then compare it when we want to use them? That way we would be 100% sure that the projects/clients belong to the current user!

I'm not explicitly requesting changes, just commenting to raise this idea... What do you think?

src/scripts/background.js
EternallLight: 🙈

EternallLight:
@alexcr I slept on all this and here's what most likely happened. Initially, I thought that I was logged out from Staging because of daily database updates - the database was updated overnight and my session got invalidated. However, with all the frequency of receiving 500 error for the large user, most likely I got this 500 that time instead. The staging server is far less powerful than prod, so retrieving user data with 14k+ projects is painful. So, this is rather an edge case.

If we add user id check, what are we going to do on mismatch? Logout the user? Re-request `me`? I think we should deal with this at some point, but not as a part of this issue (here it's about storage type change and we didn't check this for localStorage anyway), and allocate for time for investigation once we're ready.

alexcr:
@EternallLight 

> I think we should deal with this at some point, but not as a part of this issue

Sure, I agree this is not a regression so we could deal with it in another issue. However, since we're already changing this, I also think it could be trivially fixed with a couple of tweaks in your changes (see my suggestion below). Anyway, I'll approve the PR and leave this up to you 👍🏻 

> If we add user id check, what are we going to do on mismatch?

I think it should behave exactly as if the data wasn't there, and the easiest way to do that is by storing the user ID in the key itself. I've never used IndexedDB, but this is a common approach with localStorage and I assume it works with IDB too! For example:
```javascript
// setLocal:
return setIDBItem(`${TogglButton.$user.id}.${key}`, value);

// getLocal:
const idbData = await getIDBItem(`${TogglButton.$user.id}.${key}`);
```

That should be enough, because we'll store projects for user 111 under `111.projects`, and then if user 222 logs in and for some random reason we didn't clear the DB, we won't use the wrong projects because we'll try to get `222.projects` which doesn't exist.

The only potential problem is that `TogglButton.$user.id` might not be set yet when these functions are called - I haven't tested it. If that's the case, then I agree this gets more complicated and needs a follow-up issue 😅

alexcr (approved):
LGTM!! Just see my suggestion above but that's up to you :)

EternallLight:
@alexcr  I've pushed the suggested changes - that seem to work perfectly with a few line moved upwards.
Please double-check and let me know if anything is failing for you

alexcr:
@EternallLight There was a bug caused by my suggestion (sorry!) - apparently the `getLocal` function can be called from elsewhere (to get other keys - not projects or clients) in which case the user wasn't set yet. Also, I think not all keys are supposed to be per-user.

I've pushed a commit that fixed the error I saw by moving the User ID up in the call stack so that it's only used for projects and clients, so all other keys are unchanged. (I also changed the `.` separator to `-` because I noticed [we were already using this same technique elsewhere](https://github.com/toggl/toggl-button/blob/9ed981477c17d6bd500e325f83e3c1448e7f7576/src/scripts/lib/db.js#L142) but with `-` as separator, so I changed ours for consistency :)

alexcr (approved):
LGTM! :)

dpsxp (approved):
Can confirm that with the user in the original issue I got stuck at the login screen and on this version, I can log in just fine. 🚀 

Logged with another user to see if we would leak data and everything is working as expected 👍

tiberiusuciu (approved):
Tried on my own account and it works as intended on Chrome and Firefox 👍 
LGTM 🚀

PR #1999, opened 2021-08-24
fix(style): Make dark scroll bar in Chromium-based browsers

rakleed:

Make dark scroll bar in Chromium-based browsers, since the `"chrome_style": true` in the manifest does not work for the scroll bar.

Turn on the dark theme on the system and open the extension options. And also greatly reduce the size of the browser and open a pop-up window so that a scroll bar appears in it.

### DIFF
--- a/src/styles/colours.css
+++ b/src/styles/colours.css
@@ -1,4 +1,6 @@
 :root {
+  color-scheme: light dark;
+
   --white: #fff;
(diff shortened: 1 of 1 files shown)


rakleed:
@LulzAugusto @Phenome please take a look.

rakleed:
@Phenome @dpsxp @nunofmn please take a look.

PR #2013, opened 2021-09-28
fix(idb): Fix IDB initialization

### LINKED ISSUE #2004: Projects missing from extension drop down
* **OS version**: Windows 10
* **Browser version**: 93.0.4577.63
* **Extension version**: 1.70.4

### 🐛 Describe the bug

One of our users reached out to let us know that she cannot see any projects from the Browser Extension project drop down, she’s on 1.70.4, we have tried to re-sync and also verified that she’s using the correct account, funny thing is, she can see her previous entries on the extension pop-up and these show without entries even though the timer page and Bo shows they do have projects selected.

### Expected behaviour

User should see their projects on the dropdown

### Other details or context

[Slack](https://toggl.slack.com/archives/C9LMUPD9T/p1628728372273200)

Phenome:
Makes sure we "upgrade/initialize" the database before we try to interact with it.
Before this there could be a race condition where we try before we initialize, and then after this the upgrade callback never tries to run again, as the database is already on the same requested version. So whoever got to that state wouldn't be able to get out of it, only if uninstalling and reinstalling, to clear related IDB db, I believe.
I bumped the version number to make sure the upgrade code runs.

* Go to chrome://extensions and inspect background.js of the extension
* Go to the "Application" tab and delete the IndexedDB database for the extension
* Disable and re-enable the extension (sometimes this wasn't enough for me. But a `chrome://restart` always seemed to do the trick
* You'll notice console errors about accessing an object store. Also, the `LocalData` object store will never get created.
* Hopefully on this branch you won't be able to get to this state.

Closes #2004

### DIFF
--- a/src/scripts/lib/dbUtils.ts
+++ b/src/scripts/lib/dbUtils.ts
@@ -1,63 +1,66 @@
 const indexedDB = window.indexedDB;
 
-const DB_NAME = 'TogglExtension';
-const DB_VERSION = 1;
+const DB_NAME = "TogglExtension";
+const DB_VERSION = 2;
 
-const open = indexedDB.open(DB_NAME, DB_VERSION);
+const openDb = (): Promise<IDBDatabase> =>
+  new Promise((resolve, reject) => {
+    const open = indexedDB.open(DB_NAME, DB_VERSION);
 
-open.onupgradeneeded = () => {
-  const db = open.result;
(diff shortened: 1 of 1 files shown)


EternallLight (approved):
The changes look good – no more uncaught errors!
Upgrade from 1 to 2 went well.
Couldn't break it with the provided steps.
LGTM! 🚀

Issue #2002, opened 2021-09-08
Toggl Button location changed in Jira


VeljkoTN:
* **OS version**: Windows 10
* **Browser version**: Chrome 93.0.4577.63
* **Extension version**: 1.70.4

#### Relevant integration (if any):

Jira

### 🐛 Describe the bug

The button is below the "path" of an issue and should be on its right side, here's an error showing where it is and where it should be:

<img width="1128" alt="Screenshot 2021-09-06 at 15 05 26" src="https://user-images.githubusercontent.com/36152180/132538436-294777ee-c48d-42cf-a511-ab6eaf33e563.png">

### Expected behaviour

The button should be on the right

### Steps to reproduce

Steps to reproduce the behaviour:
1. Go to Jira
2. Enable the button integration
3. Go to an issue and see where the button is located.

VeljkoTN:
This is still the case with the placement of the Track browser extension: 

<img width="414" alt="Screenshot 2022-01-25 at 10 01 43" src="https://user-images.githubusercontent.com/36152180/150945387-c2ae0399-461f-4aaf-a74f-56834ce24ee6.png">

It's below the issue "path" above the title, instead of being in the same line as the "path" next to the first line, like this:

![image](https://user-images.githubusercontent.com/36152180/150945701-a9355e04-a84b-4706-b1e5-5a0ab3b7eb4f.png)

PR #2049, opened 2022-02-11
fix(jira): Fix Jira placement on Issues

dpsxp:
image

👍 

- Go to Jira
- Open the Issue
- Check the button placement

#2002


src/content/atlassian.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * @name Atlassian
```
nunofmn, line 2: Maybe it's better to keep the Atlassian / Jira naming? So that people know that it's for Jira. 😅

src/origins.js
```diff
@@ -24,7 +24,7 @@ export default {
   },
   'atlassian.com': {
     url: '*://*.atlassian.com/*',
```
nunofmn, line 26: There is also the 'atlassian.net' entry below 🤔 
Not sure if it makes sense to add both origins to the file header, or keep it like this. 🤷‍♂️

nunofmn (changes requested):
Overall, LGTM! 👍 🚀 

I think it's better to update the extension name to 'Atlassian / Jira' before merging it in though. 😅

src/origins.js
dpsxp: Looks like the .net will redirect to .com

nunofmn (approved):
LGTM! 👍 🚀

PR #2069, opened 2022-03-31
fix(notion): Add extracting description from breadcrumbs as a fallback

EternallLight:

The root cause of the issue was in missing `title` attribute on the button. The title is set during the TB initalization based on the description derived from the page content.

In Notion description is extracted from the page main section. There are cases where Toggl Button renders before the main section and it cannot get the description during the initialization. 

The same title is available in the breadcrumbs that are rendered along with the Toggl Button is the top bar. As a fallback, I added extracting the description from the breadcrumbs last text value.

Go to [this page](https://www.notion.so/toggl/Support-Team-2b9bb02020ed4efa891ec96d0cceafdb) and run the timer.

Once the popdown is hidden, the button text should be "Stop timer".

Closes https://github.com/toggl/track-extension/issues/2068

### DIFF
--- a/src/content/notion.js
+++ b/src/content/notion.js
@@ -1,3 +1,8 @@
+/**
+ * @name Notion
+ * @urlAlias notion.so
+ * @urlRegex *://*.notion.so/*
+ */
 'use strict';
 
 function createWrapper (link) {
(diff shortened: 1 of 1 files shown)


tiberiusuciu (approved):
> Once the popdown is hidden, the button text should be "Stop button".

You probably meant `Stop timer`, if so, everything is working good now 👍 

image

PR #2070, opened 2022-04-01
Add linear integration

HartS:
Add an integration with Linear (project management / issue tracking tool). This adds a toggl track button in the board view only, and inserts the issue title, project name (as project), and issue ID + labels (as tags), when the button is clicked.

Right now they don't have semantic markup which results in some pretty nasty selectors.

Debugging showed the expected tags were getting sent to the `togglbutton.createTimerLink` options, but the "Toggl tags" aren't getting selected in the time entry that gets started (even if those tags have already been created). Perhaps this is an issue with the Toggl extension itself?

I asked the Linear team to add some semantic classes, `data-*` attributes, or even `aria-` roles that could be used to construct less brittle selectors, so I'll update this PR (or submit a new one) to improve things if they're willing to oblige.

### DIFF
--- a/src/content/linear.js
+++ b/src/content/linear.js
@@ -1,0 +1,34 @@
+'use strict';
+/**
+ * @name Linear
+ * @urlAlias linear.app
+ * @urlRegex *://linear.app/*
+ */
+
+// Add linear integration in board view only
+togglbutton.render(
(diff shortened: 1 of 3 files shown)


rhys-vdw:
Is there any way this could be merged?

with-shrey:
Thanks for the contribution 👍🏼 🙏🏼

PR #2157, opened 2023-01-19
Spelling: Rename infos to info

glensc:
The noun info is uncountable.
The plural form of info is also info.

file came with initial commit: f78590c447f30eaa3685d582a2ac888302bd8f41


with-shrey:
Closed as not necessary

PR #2158, opened 2023-01-19
Fix: Add header to gitlab.js

glensc:

Fixes error reported by CI:
- https://github.com/toggl/track-extension/pull/2156#issuecomment-1397154489

reverse-engineered update-origins.js to understand what the hell the error means, as there's definitely nothing written about this in [docs/CONTRIBUTING.md](https://github.com/toggl/track-extension/blob/v2.0.19/docs/CONTRIBUTING.md)

### DIFF
--- a/src/origins.js
+++ b/src/origins.js
@@ -221,7 +221,8 @@ export default {
   },
   'gitlab.com': {
     url: '*://*.gitlab.com/*',
-    name: 'Gitlab'
+    name: 'Gitlab',
+    file: 'gitlab.js'
   },
   'gogs.io': {
     url: '*://*.gogs.io/*',
--- a/src/content/gitlab.js
+++ b/src/content/gitlab.js
@@ -1,3 +1,8 @@
+/**
+ * @name Gitlab
+ * @urlAlias gitlab.com
+ * @urlRegex *://*.gitlab.com/*
+ */
 'use strict';
 
 togglbutton.render(


src/origins.js
```diff
@@ -221,7 +221,8 @@ export default {
   },
   'gitlab.com': {
     url: '*://*.gitlab.com/*',
-    name: 'Gitlab'
+    name: 'Gitlab',
+    file: 'gitlab.js'
```
glensc, line 225: this line was automatically added by update-origins.js script.

glensc:
Should be rather trivial to write a script that:
1. takes current origins.js
2. assembles expected comment block for each integration
3. writes that to each integration replacing existing block if it's there.
4. `git add -p` to verify each change
5. commit the changes

but given past slow and negative experience with this repo, I would not do that unless someone explicitly asks to do that.

also, the error that the script outputs, should guide the user to go to `docs/CONTRIBUTING.md#file-header`, and that `# File header` block should be written first.

glensc:
Also, CI for update-origins is set up incorrectly. it runs in the context of the contributing user repository:
- https://github.com/glensc/toggl-button/actions/runs/3960074160

so, it doesn't appear at all under pull request:
- https://github.com/toggl/track-extension/pull/2158/checks

probably another reason why other integrations lack the header. I just happened to read my email to see the CI failure.

PR #2207, opened 2023-06-01
Add support for non-english zendesk

davidlienhard:

We are using the Zendesk UI from 2021. The Toggl integration is working, but only fetching the ticket number and not the name. But the title would be really helpful for us.
Currently Toggl is searching for an element with aria-label Subject. But this does not exist in the German version of Zendesk, as is named Betreff.
This PR fetches the language of Zendesk from the `<html>` node, reads the correct Label-Name from the translation array (which can be extended for other languages) and the reads the title.

You would need to test the Toggle integration in Zendesk with english and german language. Toggle schould add the ticket id and subject for tracking.

### DIFF
--- a/src/content/zendesk.js
+++ b/src/content/zendesk.js
@@ -20,14 +20,14 @@ setTimeout(() => {
       const getDescription = () => {
         const ticketId = document.querySelector('header div[data-selected=true]').attributes['data-entity-id'].value || ''
 
-		const language = document.querySelector("html").lang || 'en'
+        const language = document.querySelector("html").lang || 'en'
 
-		const translations = {
-			'en': 'Subject',
-			'de': "Betreff"
-		}
+        const translations = {
+          'en': 'Subject',
+          'de': "Betreff"
+        }
 
-		const elementName = translations[language] || 'Subject'
+        const elementName = translations[language] || 'Subject'
         const input = elem.querySelector('input[aria-label='+elementName+']');
         const title = (input ? input.value : '').trim();
 
--- a/src/content/zendesk.js
+++ b/src/content/zendesk.js
@@ -89,6 +89,7 @@ setTimeout(() => {
     function (elem) {
       let description;
       const projectName = $('title').textContent;
+
       const titleFunc = function () {
         const titleElem = $('.editable .ember-view input', elem);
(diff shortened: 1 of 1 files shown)


src/content/zendesk.js
```diff
@@ -20,7 +20,15 @@ setTimeout(() => {
...
-        const input = elem.querySelector('input[aria-label=Subject]');
+        const language = document.querySelector("html").lang || 'en'
+
+        const translations = {
+          'en': 'Subject',
+          'de': "Betreff"
+        }
+
+        const elementName = translations[language] || 'Subject'
+        const input = elem.querySelector('input[aria-label='+elementName+']');
```
nunofmn, line 31: As I said in the PR's review comment, this should be a better _locale independent_ selector.
It uses the `data-test-id` attribute that usually it's pretty stable, since it generally is used internally for E2E tests. 

```suggestion
        const input = elem.querySelector('input[data-test-id="omni-header-subject"]')
```

nunofmn (changes requested):
Hey @davidlienhard! 👋 
Thank you for submitting a PR fixing the Zendesk integration. 🙌 

Your solution is working fine, but I think we can come up with a more stable selector that is locale independent.
The last time we fixed the selector, we thought that using a selector aiming at the `aria-label` would guarantee a stable selector, that would _resist_ any potential UI changes.
But we forgot to take into consideration that the `aria-label` changes according to the user's locale. 😓 

I added a comment suggesting a change to the selector, that makes is locale independent.
If you are able to make the change, I'm happy to re-review and then approve it! 🙌 

Thank you for your contribution!

davidlienhard:
Hi @nunofmn 
Thanks for your quick feedback. I just did some tests with `data-test-id=omni-header-subject` and they have been successful. I cannot tell how stable that selector really is, but I expect it but be as stable as the old one, if not better.

Please take a look and let me know if changes are required.

Best regards, David

nunofmn (approved):
Perfect, LGTM! 👌 🚀 

Thank you for your contribution @davidlienhard! 🙌

davidlienhard:
Great, thank you for the merge. Looking forward to the next release!

Issue #2209, opened 2023-06-02
Button not showing in Gitea v1.18


duncanawoods:
OS: MacOS Ventura 13.3.1
Browser: Firefox 113.0.1 (64-bit)

The start timer button does not appear on the page. There is a "toggl" class added to an element but no element for the button. There are no errors in the console.

When debugging the extension, there are two problem:
- the issue id appears after the title not before it as the code expects
- HTMLCollection of tags does not support `forEach`

PR #2210, opened 2023-06-02
fix gitea selectors for v1.18

duncanawoods:

This PR fixes the Gitea integration for newer versions for Gitea which have reformatted the issue page.

When debugging the original code, there are two problems:

- the issue id appears after the title not before it as the code expects
- HTMLCollection of tags does not support forEach

These changes have been made.

These are fixes for Gitea v1.18. I will test against the latest version v1.19 once we update.

#2209

### DIFF
--- a/src/content/gitea.js
+++ b/src/content/gitea.js
@@ -16,7 +16,7 @@ togglbutton.render('.time-desc:not(.toggl)',
 
 function descriptionSelector () {
   const description = document.getElementById('issue-title');
-  const issueId = description.previousElementSibling;
+  const issueId = description.nextElementSibling;
   return `${issueId.textContent} ${description.textContent}`;
 }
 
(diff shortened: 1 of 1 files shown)


duncanawoods:
Confirmed that it works for latest version of gitea v19.

Only issue is that given a list of tags, they do not appear in the toggle "start task" dialogue. I can confirm in a debugger that the selector produces an array of strings.

Alissonsz (changes requested):
Hey! Thanks for your contribution.

I noticed that the description is getting a lot of blank spaces, could you fix that? Just a `trim()` should solve it.
image

duncanawoods:
Hi @Alissonsz thank you for the feedback. I couldn't replicate the excess whitespace but I have trimmed the parsed text anyway. Seems like a decent change to make. What OS/browser were you using?

Alissonsz (approved):
That's working fine now!! 
Thank you @duncanawoods, I'm testing it on Mac/Chrome.

PR #2247, opened 2023-10-09
Fixes  #2246 -  ClickUp integration the task ID is detected incorrectly

### LINKED ISSUE #2246: In ClickUp integration the task ID is detected incorrectly
* **OS version**: Windows 11
* **Browser version**: Edge latest
* **Extension version**: 3.0.18

I'm using the integration with ClickUp on public domain https://app.clickup.com

### 🐛 Describe the bug

Lately ClickUp rolled out new interface and after that the task id is not captured correctly.
Instead of the task id the integration adds null.

### Expected behaviour

The integration should capture task title and id from a page. 

### Steps to reproduce

 Press the Toggl button. 

1. Switch to UI of ClickUp to version 3.0
2. Open any task in ClickUp and press the Toggle button
3. In the track edit popup you will see the task title and the #null next to the title

### Other details or context

I checked the source code of the ClickUp integration and it looks like there is an issue there or at least the room for improvement.
In the previous commit the author takes the title tag and splits its content by a pipe "|".
Then it uses the [0] element as a track title but does nothing with the [1] element that contains the required task ID.

const description = document.querySelector('title').textContent.split('|')[0];

Instead of using the data we already have code tries to get another element that contains the task ID. And this fails in the latest version of ClickUp. Something like that:
const taskID = document.querySelector('title').textContent.split('|')[1];

As I remember, the title tag included task title and ID all the time, so it's safe to take info from it in all versions of ClickUp.

Unfortunately, I can't provide the fix because I can't enable Developer mode in the extension and test the changes.

andresgz:

Gets the ClickUp task ID from the title instead of an unexisting ID.

Unable to test it because developer mode doesn't work

https://github.com/toggl/track-extension/issues/2246

### DIFF
--- a/src/content/clickup.js
+++ b/src/content/clickup.js
@@ -5,13 +5,13 @@ togglbutton.render('#timeTrackingItem:not(.toggl)', { observe: true }, function 
 ) {
 
   const description = document.querySelector('title').textContent.split('|')[0];
-  const taskID = document.querySelector('div[role="dialog"]').getAttribute('data-task-id');
+  const taskID = document.querySelector('title').textContent.split('|')[1];
 
   const project = elem.querySelector('.space-name').textContent;
 
   const link = togglbutton.createTimerLink({
     className: 'clickup',
-    description: `${description} - #${taskID}`,
+    description: `${description}-#${taskID}`,
     projectName: project,
     buttonType: 'minimal'
   });


Alissonsz (commented):
Hey @andresgz thanks for the contribution, and sorry for the delay in answering it.
I don't know how to try UI 3.0, do you have any tips on how I could see that?
But on my UI it's now picking the description like this:
image

PR #2303, opened 2024-05-24
fix(slack): Fix toggl button disappeared on channel toolbar

with-shrey:
Fix toggl button disappeared on channel toolbar

Screenshot 2024-05-24 at 10 52 00 AM

### DIFF
--- a/src/content/hubspot.js
+++ b/src/content/hubspot.js
@@ -1,7 +1,7 @@
 /**
  * @name Hubspot
  * @urlAlias hubspot.com
- * @urlRegex *://*.atlassian.com/*
+ * @urlRegex *://app.hubspot.com/*
  */
 'use strict';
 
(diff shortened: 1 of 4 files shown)


izontm:
@with-shrey
I've added a new commit. Feel free to ignore/improve it.
Basically, it offers:

- Better alignment on top bar, moving the button to the action area
- Better alignment on the hover area
- Remove the button from the "new message" action.

image

---
_I keep thinking it is worth it to move the button to the compose message action area, so we could be better used in the threads section._

with-shrey:
@izontm love it 💚

with-shrey:
> I keep thinking it is worth it to move the button to the compose message action area, so we could be better used in the threads section.

I can visualise the usecase 
And tried it to add it 

But found a roadblock ie we cant read thread's first message if its long possibly because of virtualized list (only visible items are rendered)

May be we can title it 
thread in channel 

and created message in channel 🤔

PR #2345, opened 2024-10-04
fix(trello): optimize render for slow networks

askides:

This PR improves the TogglButton render on Trello even in slow networks.

It does still happen sometimes that the button does not appear but only in very slow networks and with the 'Disable Cache' from the dev tools checked (which is not going to be the case 99% of the time).

Tested with network throttling at 3G speed.

Ensure that the button is working.

See the [slack thread here](https://toggl.slack.com/archives/C043ZPZQ2JC/p1727968520838579?thread_ts=1727280001.195159&cid=C043ZPZQ2JC).

### DIFF
--- a/src/content/notion.js
+++ b/src/content/notion.js
@@ -50,7 +50,7 @@ togglbutton.render(
 
 togglbutton.inject(
   {
-    node: '.notion-topbar-action-buttons',
+    node: 'main.notion-frame .notion-scroller:not(.toggl)',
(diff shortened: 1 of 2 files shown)


with-shrey:
May be use the mutation observer if its flaky 

You can pass comma spperated selectors as last argument to render function 
So it will rerender button when for example description is rendered

with-shrey (approved):

PR #2372, opened 2025-02-07
fix(redbooth): Add button on right panel task drawer

ArianKrasniqi:

This PR adds the missing button when opening tasks on right panel

1. Go to Dashboard>My Tasks and open a task
2. Toggl Track button should appear in the right panel

[Thread](https://toggl.slack.com/archives/C043ZPZQ2JC/p1738932609346419) 🧵

### DIFF
--- a/src/content/redbooth.js
+++ b/src/content/redbooth.js
@@ -2,7 +2,7 @@
 
 // Right side panel
 togglbutton.render(
-  '.js-right-pane .tb-element-big:not(.toggl)',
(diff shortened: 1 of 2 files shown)


askides (commented):
Nice.

For the origins, look at the first 5 lines of the asana integration for example.

https://github.com/toggl/track-extension/blob/master/src/content/asana.js#L1-L6

You need to add this pattern as well with the relative url, as otherwise it will break.

Don't ask me why as it was already this way. :)

askides (approved):
LGTM AMAZING ❇️

PR #2376, opened 2025-03-21
fix(google-calendar): fix edit time entry event closing event popup and details selector

sanjinkapetanovich:

**Issue**
Users reported that the "Edit Time Entry" dialog was causing the event popup to close unexpectedly in the Google Calendar integration. Additionally, I discovered that the Toggl button integration was missing in the Google Calendar detail view.

**Cause**

- Google Calendar appears to have changed how it tracks clicks outside the event popup. This change affected our "Edit Time Entry" functionality, as the click on the time entry was propagating a `pointerdown` event, which closed the Google event popup.
- The Toggl button was not rendering in the detail view because the structure of the detail view in Google Calendar had changed.

**Solution**

- Implemented a MutationObserver to track the appearance of the "Edit Time Entry" dialog. This observer intercepts the pointerdown event to prevent the Google event popup from closing. Additionally, added cleanup logic to ensure the observer is properly disconnected when no longer needed.
- Updated the selector for the detail view button integration to match the new structure of the Google Calendar detail view.
- Code Cleanup and Refactor: Refactored the code for better readability and maintainability.

Event Popup Closure:

- Open Google Calendar.
- Click on an event to open the popup.
- Start a time tracking session and click anywhere on the dialog.
- Verify that neither the dialog nor the popup closes unexpectedly.

Toggl Button in Detail View:

- Open Google Calendar.
- Click on an event and then click "Edit" to open the detail view.
- Verify that the Toggl timer button is present in the detail view.

### DIFF
--- a/src/content/google-calendar.js
+++ b/src/content/google-calendar.js
@@ -34,7 +34,6 @@ function interceptPointerdownEvents(form) {
 
     if (e.target.focus) {
       e.target.focus()
-      console.log(`Focused on`, e.target)
     }
   }
 
@@ -56,7 +55,6 @@ function initializeObserver(popupElement) {
 
       // Disconnect observer when the popup element is removed
       mutation.removedNodes.forEach((node) => {
-        console.log('Removed node:', node)
         if (node.contains(popupElement)) {
           observer.disconnect()
         }
@@ -66,7 +64,6 @@ function initializeObserver(popupElement) {
 
   // Observe the entire document body for changes, the time entry edit component is a child of the body
   observer.observe(document.body, { childList: true, subtree: true })
-  console.log('MutationObserver initialized on document body')
 }
 
 function initializePopup(elem) {
@@ -90,7 +87,6 @@ function initializePopup(elem) {
 
 function initializeDetailView(elem) {
   const closeButton = $('button[data-use-native-focus-logic]', elem)
-  console.log('closeButton', closeButton)
   const getDescription = () => {
     const titleInput = $('input[data-initial-value]', elem)
     return titleInput ? titleInput.value.trim() : ''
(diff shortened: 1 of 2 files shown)


src/origins.js
```diff
@@ -252,7 +252,8 @@ export default {
   },
   'calendar.google.com': {
     url: '*://calendar.google.com/*',
-    name: 'Google Calendar'
+    name: 'Google Calendar Modern',
```
askides, line 255: Did you change the name for a specific reason? As we use if for analytics purposes and changing it can create confusion with the data.

src/content/google-calendar.js
```diff
@@ -7,71 +12,117 @@
...
+
+function initializePopup(elem) {
+  if (elem.querySelector('.toggl-button')) {
+    return // Prevent adding duplicate buttons
   }
 
-  if (!target || target.tagName.toLowerCase() !== 'div') {
-    // Prevent adding a Toggl button to other dialogs/views, e.g. Create Event
-    return;
+  initializeObserver(elem)
```
askides, line 80: Under the hood the implementation of the [render](https://github.com/toggl/track-extension-core/blob/main/src/content/index.ts#L129) method uses a mutation observer as well, asking as a curiosity, but wouldn't be possible to just add the `interceptPointerdownEvents` as the observer is already in place on the `elem`?

askides (changes requested):
Hey ✨ please could you:

1. Squash the commits as i see most of them are not much relevant by themself.
2. Check the comment on the name changed.

The other comment is a curiosity and possibly an optimization, but could be also wrong.

src/content/google-calendar.js
sanjinkapetanovich: Yeah, I noticed that, but I intentionally offloaded some logic outside the component to keep it from turning into a bloated Swiss Army Knife. I'm prioritizing reusability and maintainability over strict SRP, which seemed like a good tradeoff for widely used logic like rendering the togglButton. Also, this solution feels a bit hacky, 🙈  so I kept it separate from the main component. But yeah, let me know if you think otherwise 🙏

src/origins.js
sanjinkapetanovich: No reason 😄 , it seamed like better name, but thank you ,wasnt aware that we are using it for analytics, will revert 🙏

src/content/google-calendar.js
askides: That's fine for now.

askides (approved):
LGTM.

PR #2429, opened 2026-06-02
fix(trello): keep timer button visible and stop breaking card cover images

sanjinkapetanovich:
### Issue
On a Trello card that has a **cover image**, the card-back timer button broke in two ways:
1. **Cover image destroyed** — reopening the card shoved the cover image into a tiny thumbnail on the left and hid the button entirely.
2. **Button hard to see / inconsistent** — over a cover the button had no background and dark text, so it blended into the image; it also didn't match Trello's native header controls.

### Cause
The button was injected via `header > div > div:first-child`. That resolves to the list-name container **only when there is no cover**. When a cover exists, Trello renders it as `header`'s first `div` (`[data-testid="card-cover"]`, a `background-image` element), so `:first-child` resolved to the **cover** itself. The renderer then forced `display: inline-flex !important` on the cover (collapsing the image to a thumbnail) and appended the button **inside** it. It worked on first open only because the cover hadn't rendered yet when the observer fired.

### Solution
- **Stable anchor:** locate the list-name container via the always-present `[data-testid="card-back-actions-button"]` → its `<ul>` → the controls bar → its first child (the list dropdown). The cover lives in a separate header child, so it is never selected, flexed, or used as an injection target.
- **Native styling:** render the card-back button as a flat Trello-style pill matching the list dropdown — `--ds-background-accent-gray-subtler` background, `--ds-text` text, 3px radius, 32px height, no shadow, with a matching hover. Applied consistently on every card (cover or not) so it always looks like a Trello header control and stays legible on top of cover images.

Note: selectors are anchored to live Trello markup captured from a covered card; couldn't run the extension against Trello locally, so please verify in-browser (test plan below).

### 1. Reopening a covered card no longer breaks the image
- **Setup:** A Trello card with an image set as its cover.
- **Steps:** Open the card → close it → open it again.
- **Expected:** The cover image renders full-size every time; it is never collapsed into a thumbnail.

### 2. Button looks native and is legible
- **Setup:** Cards both with and without a cover image.
- **Steps:** Open each card → look at the header next to the list dropdown.
- **Expected:** The timer button appears as a flat gray pill matching the "This Week" list dropdown, clearly readable in both cases, and clickable.

### 3. Typing/filtering still works
- **Setup:** Any open card.
- **Steps:** Click the timer button → type a project/tag name in the Toggl popup.
- **Expected:** Keystrokes go to the popup (not captured by Trello), and the card dialog stays open.

now you see it

### DIFF
--- a/src/content/notion.js
+++ b/src/content/notion.js
@@ -13,23 +13,68 @@ function createWrapper(link) {
   return wrapper
 }
 
+// Climb from the share button to the peek panel that owns it, so the title
+// lookup stays scoped to the peek and never grabs the title of the full page
+// rendered behind it. The panel is the nearest ancestor that contains the
+// page title (the peek body lives below the topbar inside that same panel).
+function findPeekRoot(node) {
+  let current = node.closest('.notion-peek-renderer')
+  if (current) return current
+  current = node
+  while (current && current !== document.body) {
+    if (
+      current.querySelector(
+        'h1[contenteditable], h1[aria-roledescription="page title"]',
+      )
+    ) {
+      return current
+    }
+    current = current.parentElement
+  }
+  return null
+}
+
+// Find the direct child of the action-button row that holds the share button.
+// The row is the nearest ancestor that also contains the other topbar action
+// buttons; inserting the timer button before that child drops it into the row
+// rather than into the share button's own (tiny) wrapper.
+function getShareRowChild(shareButton) {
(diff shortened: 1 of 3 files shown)


sanjinkapetanovich:
Superseded by #2430, which contains the Trello fix only (this branch was accidentally based on the Notion branch and included unrelated notion.js changes).

PR #2430, opened 2026-06-02
fix(trello): keep timer button visible over card covers and match list-dropdown style

sanjinkapetanovich:
### Issue
On a Trello card that has a **cover image**, the card-back timer button broke:
1. **Cover image destroyed** — reopening the card collapsed the cover into a tiny thumbnail on the left and hid the button entirely.
2. **Button looked off** — it had no background and dark text (invisible over the cover), and didn't match Trello's native header controls.

### Cause
The button was injected via `header > div > div:first-child`. That resolves to the list-name container **only when there is no cover**. When a cover exists, Trello renders it as `header`'s first `div` (`[data-testid="card-cover"]`, a `background-image` element), so `:first-child` resolved to the **cover** itself. The renderer then forced `display: inline-flex !important` on the cover (collapsing the image) and appended the button **inside** it. It worked on first open only because the cover hadn't rendered yet when the observer fired.

### Solution
- **Stable anchor (fixes the broken cover):** locate the list-name container via the always-present `[data-testid="card-back-actions-button"]` → its `<ul>` → the controls bar → its first child (the list dropdown). The cover lives in a separate header child, so it is never selected, flexed, or used as an injection target.
- **Native styling:** render the card-back button as a flat Trello-style pill matching the **This Week** list dropdown — `--ds-background-accent-gray-subtler` background, `--ds-text` colour, `--ds-radius-small` corners, 24px height, 14px text, no shadow, with a matching hover. Custom properties are inherited from Trello's themed root, so colours/radius track Trello (including dark mode). Margins are zeroed and a 2px offset is applied so it sits flush with the dropdown. Shown consistently on every card so it always looks like a Trello control and stays legible on covers.

_Selectors are anchored to live Trello markup captured from a covered card; verified visually in-browser (alignment/size/radius matched against the list dropdown via DevTools)._

### 1. Reopening a covered card no longer breaks the image
- **Setup:** A Trello card with an image cover.
- **Steps:** Open the card → close it → open it again.
- **Expected:** The cover image renders full-size every time; it is never collapsed into a thumbnail.

### 2. Button looks native, legible, and aligned
- **Setup:** Cards both with and without a cover image.
- **Steps:** Open each card → look at the header next to the list dropdown.
- **Expected:** The timer button is a flat gray pill matching the "This Week" dropdown (same height, corner radius, vertical alignment), readable in both cases, and clickable.

### 3. Typing/filtering still works
- **Setup:** Any open card.
- **Steps:** Click the timer button → type a project/tag name in the Toggl popup.
- **Expected:** Keystrokes go to the popup (not captured by Trello), and the card dialog stays open.

pixel perfect

### DIFF
--- a/src/content/trello.js
+++ b/src/content/trello.js
@@ -13,7 +13,9 @@ const getProject = () => {
 
 const getCardName = () => {
   return (
-    document.querySelector('[data-testid="card-back-title-input"]')?.value?.trim() ?? ''
+    document
+      .querySelector('[data-testid="card-back-title-input"]')
+      ?.value?.trim() ?? ''
   )
 }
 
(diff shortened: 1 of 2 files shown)


izontm (approved):
This makes realize how easy the new extension is 😅

PR #2432, opened 2026-06-15
fix(notion): anchor side-peek timer button on close button, not shell class

sanjinkapetanovich:
### Issue
The Toggl timer button disappeared from Notion's **side-peek** view — users with the Notion integration enabled saw no button next to "Share" when opening a page in the peek. It also never appeared at all on Notion's newer **`app.notion.com`** domain.

### Cause
Two separate regressions:
1. **The peek selector broke.** Notion redesigned the side-peek topbar and moved the Share button out of the wrapper classes the integration keyed on (`.notion-peek-renderer` / `.peek-top-hover-area`). The render selector matched nothing, so the button never rendered.
2. **The newer domain was unsupported.** Notion now serves the app on `app.notion.com` (root domain `notion.com`), which the integration never registered — only `notion.so` existed in the origins, so nothing injected there.

### Solution
1. **Anchor on `.notion-peek-close`** instead of chasing Notion's rotating shell classes. The close button exists only in a peek topbar (the full-page topbar has none), so it both locates the peek's topbar and distinguishes the peek's share menu from the full-page one rendered behind it. Added an idempotency guard (skip if a wrapper already exists) and kept the existing `getShareRowChild` insertion.
2. **Added a `notion.com` origin that aliases `notion.so`** (`aliasOf: 'notion.so'`), plus the `aliasOf` field on the origin type. Aliasing means `app.notion.com` is treated as the *same* "Notion" integration — no second toggle, and existing Notion-enabled users get it automatically. The consuming alias logic lives in **toggl/track-extension-core#1547** (required companion).

### 1. Peek button restored (notion.so)
- **Setup:** Notion integration enabled; a Notion page open on `notion.so`.
- **Steps:** Open the page in side-peek → look at the topbar, left of "Share."
- **Expected:** The Toggl timer button appears; clicking it starts a timer with the page title as the description.

### 2. Works on app.notion.com
- **Setup:** Same user, Notion already enabled (no extra setup).
- **Steps:** Open a page on `app.notion.com` in side-peek.
- **Expected:** Button appears automatically — no new toggle, no re-enabling.

### 3. No full-page false positive
- **Setup:** Notion integration enabled.
- **Steps:** Open a full (non-peek) Notion page.
- **Expected:** The peek button is not injected into the full-page topbar; only the side-peek gets the close-anchored button.

> Requires **toggl/track-extension-core#1547** to be released for `app.notion.com` injection to take effect end to end.

_When the button finally shows up in the peek again:_ [reappearing magic button](https://giphy.com/explore/magic-button)

### DIFF
--- a/.github/workflows/update-origins.js
+++ b/.github/workflows/update-origins.js
@@ -65,7 +65,9 @@ exec(`git diff --name-only HEAD^ HEAD`, (_, out) => {
   const newOrigins = infos.reduce(
     (_origins, { name, urlAlias, urlRegex, file }) => ({
       ...origins,
-      [urlAlias]: { url: urlRegex, name, file },
+      // Preserve manually-maintained fields on an existing entry (e.g.
+      // `aliasOf`) instead of overwriting it with only url/name/file.
+      [urlAlias]: { ...origins[urlAlias], url: urlRegex, name, file },
     }),
     origins
   );
--- a/src/content/notion.js
+++ b/src/content/notion.js
@@ -58,16 +58,32 @@ function getShareRowChild(shareButton) {
   return child
 }
 
-// Button renders in popup/dialog (side-peek) view.
-// Target the share menu inside the peek topbar so that when React re-renders
-// and recreates the share button, the observer re-triggers. Notion's peek
-// topbar markup changes often; match both the legacy `.notion-peek-renderer`
-// shell and the newer `.peek-top-hover-area` one.
+// Walk up from the share button to the peek topbar that owns it. The topbar is
+// identified by the close button (`.notion-peek-close`), which only exists in a
+// peek — the full-page topbar has none — so this also tells a peek share menu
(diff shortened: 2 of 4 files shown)


sanjinkapetanovich:
Addressed both review points in `4ada3908`:

- **`aliasOf` stripped by the origins generator:** fixed `update-origins.js` to preserve manually-maintained fields (`[urlAlias]: { ...origins[urlAlias], url, name, file }`) and re-added `aliasOf: 'notion.so'` to the `notion.com` entry. Verified by running the generator locally — `aliasOf` now survives regeneration, and the `update-origins` CI run on this push produced no diff.
- **Peek guard could select the wrong topbar:** `getPeekTopbar` is now bounded — it resolves the action-button group's container and only accepts it if *that* container holds `.notion-peek-close`, rather than searching broad ancestors' subtrees. A full-page Share button can no longer match a peek open elsewhere in the document.

PR #2434, opened 2026-06-16
feat(outlook): add timer button to calendar event editor

sanjinkapetanovich:
### Requirement
The Outlook integration only ever injected a timer button into **inbox emails** and the **email compose** view — there was no support for **Outlook Calendar events**. Users on the calendar couldn't start a timer for an event (the only button they sometimes saw was the inbox-panel button incidentally matching the calendar page, which varies by tenant/layout).

### Solution
Add a timer button to the **calendar event editor** (create/edit), anchored on the event command toolbar, using the event title as the timer description. Two Outlook-specific wrinkles handled:
- **Self-healing injection** — Outlook re-renders the toolbar after the timer starts, dropping the button. Anchoring with `:not(:has(.toggl-button))` makes the observer re-add it the moment it's missing, so it survives re-renders.
- **Popup portal** — the post-start edit popup defaults to `document.body`, which renders *behind* Outlook's modal (z-index + focus trap) and is invisible. Portaling it into the modal (`container: '.ms-Modal-scrollableContent'`) makes it visible, same pattern as `placker.js`.

### Implementation
A new `togglbutton.render` block in `src/content/outlook.js`:
- **Anchor:** `[role="toolbar"][data-app-section="Toolbar"]:not(:has(.toggl-button))` — the event command toolbar; the `:not(:has(...))` self-heals across re-renders.
- **Scope:** only injects when a `[id*="CalendarCompose"][id$="_SUBJECT"]` subject block is present — that id is stable, not localised, and unique to the event editor (so it doesn't fire on email toolbars).
- **Description:** the event title `<input>` inside the subject block.
- **`container: '.ms-Modal-scrollableContent'`** so the edit popup renders inside the event modal.

Note: uses CSS `:has()` (Chrome/Edge 105+, Firefox 121+), which is within the extension's supported browser baseline.

### 1. Button appears in the event editor
- **Setup:** Outlook integration enabled on `outlook.office.com` (also `outlook.live.com` / `outlook.cloud.microsoft`); open Calendar.
- **Steps:** Create or open an event to bring up the editor.
- **Expected:** A Toggl timer button appears in the event command toolbar (next to Event/Series/Delete…).

### 2. Description = event title; popup shows in the modal
- **Steps:** Give the event a title, click the Toggl button to start a timer.
- **Expected:** The post-start edit popup appears **inside the modal** (visible, not hidden behind it), prefilled with the event title.

### 3. Survives Outlook's re-render
- **Steps:** Start a timer, let the popup close (submit/cancel).
- **Expected:** The toolbar button returns and remains — it does **not** stay hidden until the dialog is reopened.

### 4. Doesn't leak into email views
- **Steps:** Open an inbox email and the email compose view.
- **Expected:** The existing email/compose buttons are unchanged; the calendar block doesn't inject there (no CalendarCompose subject block).

_Tracking time straight from the calendar:_ [calendar time](https://giphy.com/explore/calendar)

### DIFF
--- a/src/content/outlook.js
+++ b/src/content/outlook.js
@@ -72,12 +72,20 @@ togglbutton.render(
   '[role="toolbar"][data-app-section="Toolbar"]:not(:has(.toggl-button))',
   { observe: true },
   (toolbar) => {
-    const dialog = toolbar.closest(".ms-Modal-scrollableContent") || document;
+    // The toolbar must live inside an event modal. Without this guard a non-modal
+    // Outlook toolbar would fall back to a document-wide lookup and could match a
+    // CalendarCompose subject from a different editor — injecting a calendar
+    // button (and the wrong event title) onto an unrelated toolbar.
+    const dialog = toolbar.closest(".ms-Modal-scrollableContent");
+    if (!dialog) {
+      return;
+    }
+
+    // Scope the subject lookup to this modal; only the calendar event editor has
+    // a CalendarCompose subject block.
     const subjectBlock = dialog.querySelector(
       '[id*="CalendarCompose"][id$="_SUBJECT"]'
     );
-
-    // Only the calendar event editor has a CalendarCompose subject block.
     if (!subjectBlock) {
       return;
     }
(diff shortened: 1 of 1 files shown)


sanjinkapetanovich:
Addressed the major:

- **Modal scope:** removed the `|| document` fallback — the toolbar must now be inside `.ms-Modal-scrollableContent`, and the `CalendarCompose` subject lookup is scoped to that modal. A non-modal Outlook toolbar can no longer match another editor's subject.
- **Popup target:** the modal is tagged with a unique `data-toggl-popup-host` (the per-instance subject id), and the popup `container` now targets `[data-toggl-popup-host="…"]` — so with multiple editors open the popup portals into *this* one, not the first modal on the page.

PR #2439, opened 2026-07-02
fix(github): strip injected 'Edit issue title' label from tracked description

### LINKED ISSUE #2438: GitHub integration prefixes issue titles with 'Edit Issue Title'
* **OS version**: Ubuntu 24.04.3 LTS
* **Browser version**: Chrome 143.0.7499.109
* **Extension version**: 4.11.16

This is happening in the github integration

### 🐛 Describe the bug

When starting a timer from a ticket in github, it creates the toggle time entry with a "Edit Issue Title" prefix in the title of the toggle entry. When saving, this prefix is passed on to toggle, so also there the problem remains.

### Expected behaviour

Expected a clean issue title without the prefix.

### Steps to reproduce

1. Go to the ticket you want to time track
2. Start the timer
3. See that the "Edit Issue Title" prefix is being added

timtebeek:
## Problem

For users with write access, GitHub renders an inline "Edit issue title" button inside the issue title element. Its visually-hidden label leaked into `titleElem.textContent` and was prepended to the Toggl time entry description (e.g. `#2438Edit issue title My title`).

## Fix

- Read the title from the dedicated `.markdown-title`/`bdi` node, with a fallback that clones the element and strips interactive/hidden nodes (`button`, `[role="button"]`, `[class*="VisuallyHidden"]`, `.sr-only`).
- Match the issue number by shape (`#\d+`) rather than DOM position, so the inline-edit wrapper span isn't mistaken for the number span.

- Fixes #2438

### DIFF
--- a/src/content/github.js
+++ b/src/content/github.js
@@ -5,6 +5,54 @@
  */
 'use strict'
 
+// For users with write access, GitHub renders an inline "Edit issue title"
+// button inside the title element. Its (visually hidden) label leaks into
+// `textContent` and gets prepended to the tracked description (see #2438).
+// Read the title from the dedicated markdown-title/bdi node instead, falling
+// back to a copy of the element with interactive/hidden nodes stripped out.
+const getIssueTitleText = (titleElem) => {
+  if (!titleElem) {
+    return ''
+  }
+
+  const titleNode = titleElem.matches('.markdown-title, bdi')
+    ? titleElem
+    : titleElem.querySelector('.markdown-title, bdi')
+
+  if (titleNode) {
+    return titleNode.textContent.trim()
+  }
+
+  const clone = titleElem.cloneNode(true)
+  clone
+    .querySelectorAll(
+      'button, [role="button"], [class*="VisuallyHidden"], .sr-only',
+    )
+    .forEach((node) => node.remove())
+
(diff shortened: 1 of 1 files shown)


timtebeek:
@TheSoloJake Is this already on a review queue somewhere internally? Would love to see this make it in.

The reason I ask to see this through is because I want to follow up with a small change that builds upon this to add the project org/repo as a tag, as seems like a common theme across a number of issues for the extension:
- https://github.com/toggl/track-extension/issues/965
- https://github.com/toggl/track-extension/issues/930
- https://github.com/toggl/track-extension/issues/444
- https://github.com/toggl/track-extension/issues/356
- https://github.com/toggl/track-extension/issues/284
- https://github.com/toggl/track-extension/issues/526

All of these would like some form beyond the #number to link back to an actual GitHub issue, especially for folks that work across multiple repositories throughout a day. Having org/repo as a tag seems like a nice middle ground to support all those use cases, as the URL can be reconstructed, without clobbering the time entry description with it.

TheSoloJake:
Heya @timtebeek ✌🏽 Our extension dev has been made aware of this issue; thanks!

timtebeek:
Thanks for the help here! I no longer see this message.

